PR #1, opened 2025-12-05
Basic project scaffold with: chat completion, streaming and models endpoints from OpenAI

SantiagoDePolonia:


cmd/heavymodel/main.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Load configuration
+	cfg, err := config.Load()
+	if err != nil {
+		slog.Error("failed to load config", "error", err)
+		os.Exit(1)
+	}
+
+	// Validate API key
+	if cfg.OpenAI.APIKey == "" {
```
SantiagoDePolonia, line 25: The validation should be moved to config loader

PR #12, opened 2025-12-09
Added correct minimal golang version in readme docker run

BartZuc:
* Updated due to problem:
  * `go: go.mod requires go >= 1.24.0 (running go 1.21.13; GOTOOLCHAIN=local)`


SantiagoDePolonia:
Thank you! That's brilliant!

PR #48, opened 2026-01-26
fix: ensure MaxTokens >= budget when extended thinking is enabled

SantiagoDePolonia:
When extended thinking is enabled via the reasoning parameter, Anthropic requires max_tokens to be at least the thinking budget_tokens. This fix ensures MaxTokens is bumped to the budget value when it's lower, in both ChatRequest and ResponsesRequest conversion paths.

## Summary by CodeRabbit

* **Chores**
  * Optimized internal reasoning effort computation to reduce redundant calculations.
  * Added validation to ensure token limits meet minimum requirements for Anthropic API requests.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/internal/providers/anthropic/anthropic.go
+++ b/internal/providers/anthropic/anthropic.go
@@ -169,6 +169,14 @@ func reasoningEffortToBudgetTokens(effort string) int {
 	}
 }
 
+// logMaxTokensAdjustment logs when MaxTokens is adjusted to meet Anthropic requirements
+func logMaxTokensAdjustment(original, adjusted int, reason string) {
+	slog.Info("MaxTokens adjusted to meet Anthropic extended thinking requirements",
+		"original", original,
+		"adjusted", adjusted,
+		"reason", reason)
+}
+
 // convertToAnthropicRequest converts core.ChatRequest to Anthropic format
 func convertToAnthropicRequest(req *core.ChatRequest) *anthropicRequest {
 	anthropicReq := &anthropicRequest{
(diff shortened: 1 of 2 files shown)


internal/providers/anthropic/anthropic.go
```diff
@@ -185,9 +185,14 @@ func convertToAnthropicRequest(req *core.ChatRequest) *anthropicRequest {
 
 	// Map reasoning effort to Anthropic extended thinking
 	if req.Reasoning != nil && req.Reasoning.Effort != "" {
```
SantiagoDePolonia, line 195: @copilot, @coderabbitai Do we need a checking against - req.Reasoning.Effort not set / nil ?

internal/providers/anthropic/anthropic.go
```diff
@@ -185,9 +185,14 @@ func convertToAnthropicRequest(req *core.ChatRequest) *anthropicRequest {
...
+		budget := reasoningEffortToBudgetTokens(req.Reasoning.Effort)
 		anthropicReq.Thinking = &anthropicThinking{
 			Type:         "enabled",
-			BudgetTokens: reasoningEffortToBudgetTokens(req.Reasoning.Effort),
+			BudgetTokens: budget,
+		}
+		// Ensure MaxTokens is at least the budget tokens
+		if anthropicReq.MaxTokens < budget {
+			anthropicReq.MaxTokens = budget
```
SantiagoDePolonia, line 205: @copilot Add some logs here - that MaxTokens has been increased because ... Anthropic require etc...

PR #83, opened 2026-02-20
fix(me/gom-33): anthropic reasoning error when i request an opus model

giorgi-pro:
Two issues were fixed for Anthropic's newer 4.6 models (e.g. `claude-opus-4-6`):

1. `max_tokens` **error with reasoning** — The 4.6 models support a new adaptive thinking mode with output_config.effort instead of the legacy budget_tokens approach. The code was applying the legacy path to all models, which caused Anthropic to reject requests with `"max_tokens must be greater than thinking.budget_tokens"`. The fix adds model detection (`isAdaptiveThinkingModel`) and uses the appropriate API format per model generation.

2. **Empty response with reasoning enabled** — When adaptive thinking is active, Anthropic can return multiple content blocks: possibly a preamble text block (in our case, just `\n\n`), a thinking block, and the actual answer text block. The fix introduces extractTextContent which picks the last text block — matching Anthropic's pattern of preamble → thinking → answer.

After the fix:

```
{
  "id": "msg_019HyRtW4EDkQgSAK5ShsH1a",
  "object": "response",
  "created_at": 1771623447,
  "model": "claude-opus-4-6",
  "provider": "anthropic",
  "status": "completed",
  "output": [
    {
      "id": "msg_291c31c4-38ab-4cd0-bdbe-c3d7000901a1",
      "type": "message",
      "role": "assistant",
      "status": "completed",
      "content": [
        {
          "type": "output_text",
          "text": "# Bash Matrix Transpose Script\n\n```bash\n#!/bin/bash\n\n# Usage: ./transpose.sh \"[1,2],[3,4],[5,6]\"\n# Output: [1,3,5],[2,4,6]\n\ninput=\"$1\"\n\nif [[ -z \"$input\" ]]; then\n    echo \"Usage: $0 \\\"[1,2],[3,4],[5,6]\\\"\"\n    exit 1\nfi\n\n# ---------------------------------------------------------------------------\n# 1. Parse: turn \"[1,2],[3,4],[5,6]\" into a clean \"1,2;3,4;5,6\" form\n#    - replace ],[ with ;   (row delimiter)\n#    - strip the leading [ and trailing ]\n# ---------------------------------------------------------------------------\ncleaned=$(echo \"$input\" | sed 's/\\],\\[/;/g' | sed 's/^\\[//' | sed 's/\\]$//')\n\n# ---------------------------------------------------------------------------\n# 2. Read into a flat array that simulates a 2-D matrix\n# ---------------------------------------------------------------------------\nIFS=';' read -ra rows <<< \"$cleaned\"\nnum_rows=${#rows[@]}\nnum_cols=0\n\ndeclare -a matrix\n\nfor ((i = 0; i < num_rows; i++)); do\n    IFS=',' read -ra cols <<< \"${rows[$i]}\"\n    # Set column count from the first row\n    if (( i == 0 )); then\n        num_cols=${#cols[@]}\n    fi\n    for ((j = 0; j < ${#cols[@]}; j++)); do\n        matrix=$(( i * num_cols + j ))          # flat index\n        matrix[$((i * num_cols + j))]=\"${cols[$j]}\"\n    done\ndone\n\n# ---------------------------------------------------------------------------\n# 3. Build the transposed output\n#    Original  : num_rows × num_cols\n#    Transposed: num_cols × num_rows\n# ---------------------------------------------------------------------------\nresult=\"\"\n\nfor ((j = 0; j < num_cols; j++)); do\n    row=\"[\"\n    for ((i = 0; i < num_rows; i++)); do\n        (( i > 0 )) && row+=\",\"\n        row+=\"${matrix[$((i * num_cols + j))]}\"\n    done\n    row+=\"]\"\n\n    if [[ -n \"$result\" ]]; then\n        result+=\",$row\"\n    else\n        result=\"$row\"\n    fi\ndone\n\necho \"$result\"\n```\n\n## How it works\n\n| Step | What happens | Example value |\n|------|-------------|---------------|\n| **Input** | Raw matrix string | `[1,2],[3,4],[5,6]` |\n| **Clean** | Replace `],[` → `;`, strip outer `[]` | `1,2;3,4;5,6` |\n| **Split rows** | Split on `;` | `[\"1,2\", \"3,4\", \"5,6\"]` |\n| **Flat array** | Store elements row-major | `matrix = [1,2,3,4,5,6]` (3×2) |\n| **Transpose** | Iterate columns-first, rows-second | Swap `(i,j)` → `(j,i)` |\n| **Output** | Rebuild bracket format | `[1,3,5],[2,4,6]` |\n\n## Test runs\n\n```bash\n$ ./transpose.sh \"[1,2],[3,4],[5,6]\"\n[1,3,5],[2,4,6]\n\n$ ./transpose.sh \"[1,2,3],[4,5,6]\"\n[1,4],[2,5],[3,6]\n\n$ ./transpose.sh \"[1],[2],[3]\"\n[1,2,3]\n\n$ ./transpose.sh \"[1,2,3]\"\n[1],[2],[3]\n\n$ ./transpose.sh \"[10,20,30],[40,50,60],[70,80,90]\"\n[10,40,70],[20,50,80],[30,60,90]\n```\n\nThe key insight is that **transposing** simply means iterating **column-first, then row** instead of the original row-first order, which swaps the `(i, j)` indices to `(j, i)`."
        }
      ]
    }
  ],
  "usage": {
    "input_tokens": 61,
    "output_tokens": 1866,
    "total_tokens": 1927
  }
}
```

## Summary by CodeRabbit

* **New Features**
  * Support for adaptive "thinking" modes on newer models with configurable effort and output options.

* **Improvements**
  * More reliable extraction of assistant text from responses that include intermediate thinking blocks.
  * Centralized reasoning handling with model-aware behavior, token adjustments, and cleaner request formatting.

* **Tests**
  * Added tests covering thinking-block extraction and adaptive-model recognition.

### DIFF
--- a/internal/providers/anthropic/anthropic.go
+++ b/internal/providers/anthropic/anthropic.go
@@ -90,7 +90,7 @@ type anthropicThinking struct {
 
 // anthropicOutputConfig controls effort level for adaptive thinking on 4.6 models.
 type anthropicOutputConfig struct {
-	Effort string `json:"effort"`
+	Effort string `json:"effort,omitempty"`
 }
 
 // anthropicRequest represents the Anthropic API request format
--- a/internal/providers/anthropic/anthropic.go
+++ b/internal/providers/anthropic/anthropic.go
@@ -185,10 +185,20 @@ type anthropicModelsResponse struct {
 // applyReasoning configures thinking and effort on an anthropicRequest.
 // 4.6 models use adaptive thinking with output_config.effort.
 // Older models use manual thinking with budget_tokens.
+func normalizeEffort(effort string) string {
+	switch effort {
+	case "low", "medium", "high":
+		return effort
+	default:
+		slog.Warn("invalid reasoning effort, defaulting to 'low'", "effort", effort)
+		return "low"
+	}
+}
+
 func applyReasoning(req *anthropicRequest, model, effort string) {
 	if isAdaptiveThinkingModel(model) {
 		req.Thinking = &anthropicThinking{Type: "adaptive"}
-		req.OutputConfig = &anthropicOutputConfig{Effort: effort}
+		req.OutputConfig = &anthropicOutputConfig{Effort: normalizeEffort(effort)}
 	} else {
 		budget := reasoningEffortToBudgetTokens(effort)
 		req.Thinking = &anthropicThinking{
@@ -213,15 +223,12 @@ func applyReasoning(req *anthropicRequest, model, effort string) {
 }
 
 func reasoningEffortToBudgetTokens(effort string) int {
-	switch effort {
-	case "low":
-		return 5000
+	switch normalizeEffort(effort) {
 	case "medium":
 		return 10000
 	case "high":
 		return 20000
 	default:
-		slog.Warn("inappropriate reasoning effort, defaulting to 'low'", "effort", effort)
 		return 5000
 	}
 }
--- a/internal/providers/anthropic/anthropic.go
+++ b/internal/providers/anthropic/anthropic.go
@@ -41,10 +41,10 @@ type Provider struct {
 func New(apiKey string, opts providers.ProviderOptions) core.Provider {
 	p := &Provider{apiKey: apiKey}
 	cfg := llmclient.Config{
-		ProviderName: "anthropic",
-		BaseURL:      defaultBaseURL,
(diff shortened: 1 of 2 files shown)


SantiagoDePolonia (approved):
LGTM!

PR #92, opened 2026-02-25
Fix: remove close(l.buffer) to prevent send-on-closed-channel panic i…

SantiagoDePolonia:
…n async loggers

A race existed between Write() and flushLoop() during shutdown: a goroutine could pass the l.closed check, get preempted, and resume after close(l.buffer), causing a panic. Closing the buffer channel is unnecessary since flushLoop exits via l.done; the GC reclaims the channel. Drain now uses a non-blocking select loop instead of range.


SantiagoDePolonia:
@coderabbitai review

PR #93, opened 2026-02-25
perf: cache sorted model lists in registry to avoid per-request sorting

SantiagoDePolonia:
ListModels(), ListModelsWithProvider(), and ListModelsWithProviderByCategory() were allocating and sorting a new slice on every call. For registries with hundreds of models under heavy concurrent load, this wastes CPU and adds GC pressure. Cache sorted slices internally and rebuild lazily (double-checked locking) only when Initialize(), LoadFromCache(), or EnrichModels() modify the underlying models map.

## Summary by CodeRabbit

* **Performance Improvements**
  * Implemented intelligent caching for model listings and category filters to improve response times.

* **New Features**
  * Added asynchronous initialization support for faster startup.
  * Added automatic background refresh capability to keep model lists current at configurable intervals.


SantiagoDePolonia:
@coderabbitai  review

PR #95, opened 2026-02-25
fix: prevent bytes.Buffer capacity retention in streaming usage wrapper

SantiagoDePolonia:
Replace Reset()+Write() with right-sized buffer allocation when the event buffer grows disproportionately large (e.g. from a single large SSE event). This ensures the oversized backing array is released to GC instead of being pinned for the stream's lifetime.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal memory management for event stream processing, enhancing efficiency.


SantiagoDePolonia:
@coderabbitai review

PR #117, opened 2026-03-06
Fix: SSE flushing and responses done markers

SantiagoDePolonia:
Bug fix:
- **Streaming is broken.** SSE tokens buffer at 4KB instead of flushing incrementally (C-04). The stream termination signal `[DONE]` is lost before delivery (C-06). Users see frozen output followed by bursts. This is the defining failure mode for an AI gateway.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streamed responses now reliably end with the terminal data: [DONE] marker and better handle upstream EOFs and streaming errors to avoid truncated outputs.
  * Improved incremental delivery to clients with buffered flush behavior for smoother real-time streaming.

* **Documentation**
  * Clarified streaming semantics in Getting Started and Tips: SSE chunks flush incrementally and streams terminate with data: [DONE].

* **Tests**
  * Added tests for marker handling, chunked/split markers, flushing, timing, and streaming error scenarios.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,9 @@ jobs:
         uses: golangci/golangci-lint-action@v9
         with:
           version: v2.6
+          # Temporary workaround: config verification fetches the remote JSON schema and
+          # intermittently times out in CI before lint runs. Re-enable by 2026-04-01 once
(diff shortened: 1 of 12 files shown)


SantiagoDePolonia:
@coderabbitai review

PR #122, opened 2026-03-06
Fix: anthropic format

SantiagoDePolonia:
Do these changes help Anthropic support?

  - Yes, model_context_window_exceeded -> length is beneficial. Without it,
    OpenAI-compatible clients would see a raw Anthropic stop reason instead of
    the expected truncation-style finish reason.
  - Yes, but narrowly, the malformed-stream fallback is beneficial. If Anthropic
    says stop_reason: "tool_use" but no tool_calls chunk was ever emitted,
    returning raw tool_use is more honest than fabricating OpenAI-style
    tool_calls.

  Are those arguments/fields used?

  - emittedToolCalls: yes, fully used in code. It exists only to distinguish
    “Anthropic said tool_use” from “the client actually received tool call
    chunks.”
  - model_context_window_exceeded: yes, it is a real Anthropic stop-reason
    value, documented for Claude 4.5+ and older models behind a beta header.
  - No new request-side Anthropic arguments or payload fields are introduced by
    this branch. This branch only changes response/stream interpretation plus
    tests.

  Net assessment:

  - This PR still makes sense, but only as a narrow compatibility/defensive fix.
  - The model_context_window_exceeded mapping is clearly useful.
  - The tool_use fallback is defensive rather than mainstream; it protects
    partial or malformed streams, not the normal happy path.

  Residual risk:

  - The fallback case is unit-tested in anthropic_test.go:504, but it is not
    backed by a real contract/integration capture, so its value is mainly
    robustness rather than coverage of a documented normal stream shape.

  Sources:

  - Anthropic stop reasons: https://docs.claude.com/en/api/handling-stop-reasons
  - Claude 4.5 release notes on model_context_window_exceeded:
    https://docs.claude.com/en/release-notes/overview

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced streaming response handling to properly manage tool use scenarios when no tool calls are emitted, ensuring correct behavior in edge cases.
  * Improved error handling for context window exceeded scenarios to maintain consistency.

* **Tests**
  * Added test coverage for tool use streaming scenarios and context window handling.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,10 @@ jobs:
         uses: golangci/golangci-lint-action@v9
         with:
           version: v2.6
+          # Temporary workaround: config verification fetches the remote JSON schema and
+          # intermittently times out in CI before lint runs. Re-enable by 2026-04-01 once
+          # the schema fetch is reliable again; track in this PR/branch discussion.
+          verify: false
 
   test-unit:
     name: Unit Tests
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,18 @@
 version: "2"
 linters:
+  enable:
+    - dupl
+  settings:
+    dupl:
+      threshold: 150
   exclusions:
+    rules:
+      - linters:
+          - dupl
+        path: _test\.go
+      - linters:
+          - dupl
+        path: tests/
     generated: lax
     presets:
       - comments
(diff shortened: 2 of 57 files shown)


SantiagoDePolonia:
@coderabbitai review

SantiagoDePolonia:
@coderabbitai review

PR #34, opened 2026-01-16
chore: improved good defaults

SantiagoDePolonia:
## Summary by CodeRabbit

* **Chores**
  * Body/header logging now defaults to enabled when related environment variables are unset.
  * Removed several logging-related environment variables from the service configuration.

* **Documentation**
  * Updated helper documentation to clarify that environment variables return their boolean value or a provided default when not set.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #123, opened 2026-03-06
feat(gom-48): add simple response cache

giorgi-pro:
# Refactor: Separate `internal/cache/` into layered packages

## Summary

Split the flat `internal/cache/` package into a generic key-value store layer (`internal/cache/`) and a model-specific domain layer (`internal/cache/modelcache/`). Each cache consumer now depends on the correct abstraction level. No behavioral changes.

## Problem

The original `internal/cache/` package mixed three unrelated concerns in a single namespace:

- **Generic KV store** (`Store`, `MapStore`, `RedisStore`) — used by response cache middleware
- **Model-cache domain** (`ModelCache`, `CachedProvider`, `CachedModel`, `Cache`, `LocalCache`, `redisModelCache`) — used by provider registry
- **Redis client construction** (`NewRedisStore`) — shared infrastructure

This caused:

- **Misleading naming** — `cache.go` defined both the generic `Store` interface and model-specific types. The name conveyed nothing about its actual contents.
- **Inverted dependencies** — the response cache imported `internal/cache` to get `Store` and `RedisStore`, but that package was conceptually the "model cache" package.
- **Misplaced constants** — `ResponseCachePrefix` (`"gomodel:response:"`) lived in `redis.go` next to `DefaultRedisKey` (`"gomodel:models:"`). A response-cache constant had no business in the model-cache package.
- **No path to connection sharing** — because `RedisStore` construction was tangled with model-cache logic, there was no clean seam to share a single Redis connection between consumers.

## Solution

```
internal/cache/                        ← Generic KV store layer
  store.go                             ← Store interface + MapStore
  redis.go                             ← RedisStore + DefaultRedisTTL

internal/cache/modelcache/             ← Model-cache domain (NEW)
  modelcache.go                        ← ModelCache, CachedProvider, CachedModel, Cache interface
  local.go                             ← LocalCache (file-based)
  redis.go                             ← redisModelCache (adapts cache.Store → Cache)
  local_test.go                        ← LocalCache + serialization tests
  redis_test.go                        ← redisModelCache tests via MapStore

internal/responsecache/                ← Unchanged structure
  responsecache.go                     ← Now owns its own responseCachePrefix constant
  simple.go                            ← Imports cache.Store (correct layer)
  middleware_test.go                   ← Imports cache.NewMapStore (correct layer)
```

**Dependency direction (after):**

```
responsecache  ──▶  cache              (Store, RedisStore)
modelcache     ──▶  cache              (Store, RedisStore, DefaultRedisTTL)
providers      ──▶  modelcache         (Cache, ModelCache, LocalCache, ...)
providers      ──▶  cache              (DefaultRedisTTL only)
```

Both `responsecache` and `modelcache` depend on the generic `cache` layer. Neither depends on the other.

### New files

| File | Contents | Origin |
|---|---|---|
| `internal/cache/store.go` | `Store` interface, `MapStore` | Extracted from old `cache.go` |
| `internal/cache/modelcache/modelcache.go` | `ModelCache`, `CachedProvider`, `CachedModel`, `Cache` interface | Extracted from old `cache.go` |
| `internal/cache/modelcache/local.go` | `LocalCache` file-based impl | Moved from old `local.go` |
| `internal/cache/modelcache/redis.go` | `redisModelCache`, `DefaultRedisKey` | Moved from old `model_redis.go` |
| `internal/cache/modelcache/redis_test.go` | Redis model cache tests | Moved from old `model_redis_test.go` |
| `internal/cache/modelcache/local_test.go` | Local cache + serialization tests | Moved from old `cache_test.go` |

### Deleted files

| File | Reason |
|---|---|
| `internal/cache/cache.go` | Split into `store.go` + `modelcache/modelcache.go` |
| `internal/cache/local.go` | Moved to `modelcache/local.go` |
| `internal/cache/model_redis.go` | Moved to `modelcache/redis.go` |
| `internal/cache/model_redis_test.go` | Moved to `modelcache/redis_test.go` |
| `internal/cache/cache_test.go` | Moved to `modelcache/local_test.go` |

### Modified files

| File | Change |
|---|---|
| `internal/cache/redis.go` | Removed `DefaultRedisKey` and `ResponseCachePrefix` constants |
| `internal/responsecache/responsecache.go` | Added local `responseCachePrefix` constant; removed `cache.ResponseCachePrefix` reference |
| `internal/providers/init.go` | Import `modelcache`; `cache.Cache` → `modelcache.Cache`, `cache.NewRedisModelCache` → `modelcache.NewRedisModelCache`, `cache.NewLocalCache` → `modelcache.NewLocalCache`, `cache.DefaultRedisKey` → `modelcache.DefaultRedisKey` |
| `internal/providers/registry.go` | Import `modelcache`; all `cache.ModelCache`/`cache.CachedProvider`/`cache.CachedModel` → `modelcache.*` |
| `internal/providers/registry_cache_test.go` | Same import/type changes as `registry.go` |
| `tests/stress/stress_test.go` | Import `modelcache`; `emptyCache` mock uses `modelcache.ModelCache` |

### Unchanged

- `internal/responsecache/simple.go`, `middleware_test.go` — already imported `cache.Store` correctly
- `internal/app/app.go`, `internal/server/http.go` — no direct cache type references
- `config/` — config DTOs don't depend on cache packages

## What did NOT change

- **No behavioral changes.** All interfaces, method signatures, and runtime behavior are identical.
- **No config changes.** YAML/env var structure is untouched.
- **No new dependencies.** `modelcache` imports only its parent `internal/cache` and stdlib.
- **Connection behavior unchanged.** Model cache and response cache still create separate Redis connections.

## Future: shared Redis connection

With the generic `cache.Store` layer separated, sharing a single Redis connection becomes a small localized change: create one `RedisStore` in `app.go` at startup and pass it into both `modelcache.NewRedisModelCacheWithStore(sharedStore, ...)` and `responsecache.NewResponseCacheMiddlewareWithStore(sharedStore, ...)`. Both constructor-with-store variants already exist.

## Summary by CodeRabbit

* **New Features**
  * Response caching middleware (Redis-backed with no-op fallback) to speed repeated API requests.

* **Improvements**
  * Split cache configuration into separate model and response sections with independent Redis prefixes and TTLs.
  * Validation added to ensure exactly one model cache backend is selected.
  * Response cache integrated into server lifecycle for proper startup and shutdown.
  * In-memory key-value store added for local/testing use.

* **Tests**
  * Extensive tests for cache validation, model cache backends, and response cache behavior.


config/config.go
```diff
@@ -251,41 +251,63 @@ type MongoDBStorageConfig struct {
...
 
-	// RefreshInterval is how often to refresh the model registry in seconds.
-	// Default: 3600 (1 hour)
-	RefreshInterval int `yaml:"refresh_interval" env:"CACHE_REFRESH_INTERVAL"`
+// RedisCacheConfig holds Redis connection configuration. Shared by model and response cache.
+type RedisCacheConfig struct {
+	URL string `yaml:"url" env:"REDIS_URL"`
+	Key string `yaml:"key" env:"REDIS_KEY"`
+	TTL int    `yaml:"ttl" env:"REDIS_TTL"`
+}
```
SantiagoDePolonia: This common type probably should not exist. IIUC it's used in both:
- model list cache
- simple response cache

It means that we have one env variable: `REDIS_KEY` to configure both REDIS keys.

We should have two variable for this:
- REDIS_KEY_MODELS
- REDIS_KEY_SIMPLE_RESPONSES_CACHE

or something like this.

Am I correct?

giorgi-pro: You are right. I wanted to share the types whenever it made sense, but it came with this side effect. Will apply some solution. Thanks

internal/responsecache/simple.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+type responseCapture struct {
+	http.ResponseWriter
+	body *bytes.Buffer
+}
+
+func (r *responseCapture) Write(b []byte) (int, error) {
+	r.body.Write(b)
+	return r.ResponseWriter.Write(b)
```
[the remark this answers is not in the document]
giorgi-pro: The suggested fix would break things. It would silently drop the response to the client on every cache miss. There is no other code path that copies from capture.body to the underlying writer. The comment in the suggestion ("the real writing to the client should already be happening elsewhere") is factually incorrect — reading the middleware shows that responseCapture.Write is the only write path to the client for non-cached responses. 

There is no XSS here because:

1. The Content-Type is always application/json, so browsers won't parse it as HTML
2. The bytes originate from the upstream LLM API, not from the client's request
3. This is a JSON API gateway — it has no HTML rendering surface

Adding comments to make the intent clearer.

internal/responsecache/simple.go
```diff
@@ -0,0 +1,132 @@
...
+	http.ResponseWriter
+	body *bytes.Buffer
+}
+
+func (r *responseCapture) Write(b []byte) (int, error) {
+	// Write to the underlying ResponseWriter first so the client always receives
+	// the response. Buffer a copy separately for cache storage only.
+	// Note: b originates from upstream LLM API responses (JSON), not from
+	// client-controlled input, so there is no XSS risk here.
+	n, err := r.ResponseWriter.Write(b)
```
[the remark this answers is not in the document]
giorgi-pro: Already addressed

docker-compose.yaml
```diff
@@ -11,7 +11,7 @@ services:
       # Cache configuration
       - CACHE_TYPE=redis
       - REDIS_URL=redis://redis:6379
-      - REDIS_KEY=gomodel:models
+      - REDIS_KEY_MODELS=gomodel:models
```
SantiagoDePolonia: @coderabbitai Do we need this? The default value should be enought

.env.template
```diff
@@ -31,7 +31,10 @@
 
 # Redis Configuration (only used when CACHE_TYPE=redis)
 # REDIS_URL=redis://localhost:6379
-# REDIS_KEY=gomodel:models
+# REDIS_KEY_MODELS=gomodel:models
+# REDIS_TTL_MODELS=86400
+# REDIS_KEY_RESPONSES=gomodel:response:
```
SantiagoDePolonia, line 36: @coderabbitai Why we have ":" on the end here, but not at line #34 (two lines aboce?

docker-compose.yaml
SantiagoDePolonia: @giorgi-pro you can remove this line

docs/advanced/configuration.mdx
```diff
@@ -50,8 +50,15 @@ The most common way to configure GOModel. Set any of the variables below to over
...
-| `REDIS_TTL`         | Cache TTL in seconds              | `86400` (24h)    |
+| `REDIS_KEY_MODELS`     | Redis key for model cache           | `gomodel:models`    |
+| `REDIS_KEY_RESPONSES`  | Redis key for response cache        | `gomodel:response:` |
+| `REDIS_TTL_MODELS`     | TTL in seconds for model cache      | `86400` (24h)    |
+| `REDIS_TTL_RESPONSES`  | TTL in seconds for response cache   | `3600` (1h)      |
+
+<Note>
+  **Migration note:** In previous versions, `REDIS_KEY` and `REDIS_TTL` were used for the model cache.
+  These have been renamed to `REDIS_KEY_MODELS` and `REDIS_TTL_MODELS`. Update your configuration if upgrading.
+</Note>
```
SantiagoDePolonia: @giorgi-pro We don't need this comment. Currently we don't need to care about backward-compatibility.

SantiagoDePolonia (approved):
Generally - looks good. But please change those two places where I marked you in the comment and feel free to merge

helm/values.schema.json
```diff
@@ -191,9 +191,11 @@
...
-            "url": { "type": "string" },
-            "key": { "type": "string" },
-            "ttl": { "type": "integer" }
+            "url": { "type": "string", "description": "Redis connection URL" },
+            "keyModels": { "type": "string", "description": "Redis key prefix for the model registry cache (default: gomodel:models)" },
+            "ttlModels": { "type": "integer", "description": "TTL for model cache entries in seconds (default: 86400)" },
+            "keyResponses": { "type": "string", "description": "Redis key prefix for the response cache (default: gomodel:response:)" },
+            "ttlResponses": { "type": "integer", "description": "TTL for response cache entries in seconds (default: 3600)" }
           }
         }
```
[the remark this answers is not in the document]
giorgi-pro: Already fixed. @coderabbitai verify

PR #200, opened 2026-04-01
fix(responsecache): cache streaming responses

SantiagoDePolonia:
## Summary
- cache streamed chat and responses requests by reconstructing a full canonical response for storage
- replay cached streaming hits as SSE while preserving reasoning deltas and request-specific usage behavior
- keep stream and non-stream requests in the same cache namespace without collapsing meaningful stream_options differences

## Testing
- go test ./...

## Summary by CodeRabbit

* **New Features**
  * Streaming requests can now be cached; previously they were always bypassed.
  * Cached responses can be served in alternative formats—streaming requests can receive cached JSON responses as SSE, and non-streaming requests can receive SSE-cached responses as JSON.

* **Bug Fixes**
  * `Cache-Control: no-cache` header now properly bypasses cache layers (previously only `no-store` did).

* **Improvements**
  * Cache keys now account for `stream_options.include_usage` parameter variations, ensuring correct cache segregation.

### DIFF
--- a/docs/about/who-we-are.mdx
+++ b/docs/about/who-we-are.mdx
@@ -68,4 +68,5 @@ project — not individual developers or small teams building with it.
 
 ## Links
 
+- **Enterpilot:** [enterpilot.io](https://enterpilot.io)
 - **Discord:** [Join our server](https://discord.gg/gaEB9BQSPH)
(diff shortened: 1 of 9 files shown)


SantiagoDePolonia:
@coderabbitai review

PR #206, opened 2026-04-02
feat: UI experimentation chart two rows

SantiagoDePolonia:
Example preview:

Screenshot 2026-04-03 at 00 51 14
Screenshot 2026-04-03 at 00 51 25


SantiagoDePolonia:
The current version is better.

PR #248, opened 2026-04-20
feat: improve passthrough support

giorgi-pro:
## Summary by CodeRabbit

* **New Features**
  * Passthrough routing keyed by provider instance: /p/{instance}/... with optional /p/{instance}/v1/... alias; per-instance opt-out via passthrough_disabled.

* **Refactor**
  * Passthrough now forwards request bodies and proxies upstream responses verbatim (including SSE/errors/status); semantic enrichment removed.
  * Request ID header unified to X-GoModel-Request-ID across the system.

* **Documentation**
  * Removed ENABLED_PASSTHROUGH_PROVIDERS env guidance and updated docs/examples/roadmap for instance-based passthrough.


config/config.example.yaml
```diff
@@ -8,9 +8,8 @@ server:
   master_key: "your-secret-key"
   body_size_limit: "10M"
   pprof_enabled: false # expose /debug/pprof/* for local profiling only
-  enable_passthrough_routes: true # expose /p/{provider}/{endpoint} passthrough routes
-  allow_passthrough_v1_alias: true # allow /p/{provider}/v1/... while keeping /p/{provider}/... canonical
-  enabled_passthrough_providers: ["openai", "anthropic", "openrouter", "zai", "vllm"] # providers enabled on /p/{provider}/...
```
SantiagoDePolonia, line 13: I would add an option like:

`disable_passthrough_api: false  #yaml`


And for every provider:
`passthrough_enabled: true #yaml` 

ENV variables should be like:
`DISABLE_PASSTHROUGH_API=false`
`OPENAI_PASSTHROUGH_ENABLED=true`

docs/features/passthrough-api.mdx
```diff
@@ -18,7 +18,9 @@ translating the request or response body.
 
 Use it when a client library expects a provider's native API instead of the
 OpenAI-compatible API. For example, the Anthropic SDK sends requests to
-`/v1/messages`, so you can point it at GoModel's Anthropic passthrough base URL:
+`/v1/messages`, so you can point it at GoModel's Anthropic passthrough base URL. The
+first path segment must match a **key** under `providers:` in your config (not the
```
SantiagoDePolonia, line 22: Change `key` to `provider_name (key)`. Or something like this

docs/features/passthrough-api.mdx
```diff
@@ -39,10 +41,10 @@ POST /messages
 
 ## How it works
 
-Passthrough routes use this shape:
+Passthrough routes use this shape (first segment = YAML instance name / `providers` key):
 
 ```text
-/p/{provider}/{endpoint}
+/p/{instance}/{endpoint}
```
SantiagoDePolonia, line 47: Let's stick to provider_name - it's more precise. However, we can discuss it.

docs/features/passthrough-api.mdx
```diff
@@ -39,10 +41,10 @@ POST /messages
 
 ## How it works
 
-Passthrough routes use this shape:
+Passthrough routes use this shape (first segment = YAML instance name / `providers` key):
```
SantiagoDePolonia, line 44: I would remove the parentheses here.

docs/features/passthrough-api.mdx
```diff
@@ -130,8 +132,9 @@ from passthrough requests before forwarding them upstream.
 
 Passthrough is intentionally narrow while the API is in beta.
 
-- `openai`, `anthropic`, `openrouter`, `zai`, and `vllm` are enabled by
-  default.
+- Passthrough is available for any configured **provider instance** that implements
+  it. There is no global allowlist: use `passthrough_disabled: true` on a
+  `providers.<name>` block to block `/p/<name>/...` for that instance.
```
SantiagoDePolonia, line 137: let's use <provider_name> here, even it it might be redundant here, let's do it to be consequent and to make this string easier to find across the whole codebase

docs/features/passthrough-api.mdx
```diff
@@ -145,7 +148,7 @@ Passthrough routes are enabled by default:
 ```env
 ENABLE_PASSTHROUGH_ROUTES=true
 ALLOW_PASSTHROUGH_V1_ALIAS=true
-ENABLED_PASSTHROUGH_PROVIDERS=openai,anthropic,openrouter,zai,vllm
 ```
 
-Set `ENABLED_PASSTHROUGH_PROVIDERS` to the provider types you want to expose.
+Opt out per **instance** in `config/config.yaml` with `passthrough_disabled: true`
```
SantiagoDePolonia, line 153: Generally we should be focused on ENV variables in doc and config.yaml should be optional - that's our direction. Detailed things about the config.yaml should be written at config-yaml specific docs page OR documented at `config.example.yaml`

docs/guides/vllm.mdx
```diff
@@ -125,9 +125,12 @@ GoModel handles gateway authentication, strips client auth headers before
 forwarding, and applies `VLLM_API_KEY` to upstream requests when configured.
 
 <Note>
-  Passthrough routes are provider-type scoped at `/p/vllm/...`. When you need
-  to target one named vLLM instance in a multi-provider setup, use translated
-  `/v1/...` requests with provider-qualified model IDs such as
+  The first path segment is your **provider instance name** in config (e.g. `/p/vllm/...`
```
SantiagoDePolonia, line 128: Let's change instance name -> provider_name

README.md
```diff
@@ -207,6 +206,57 @@ Key settings:
 
 ---
 
+## Provider-Native Passthrough
+
+GoModel exposes provider-native endpoints via `/p/{instance}/...` routes (the first segment is the **instance name**, i.e. the key under `providers:` in config), allowing direct access to provider APIs with minimal gateway involvement. This is useful when you need raw provider responses or specific provider features not abstracted by the OpenAI-compatible API.
```
SantiagoDePolonia, line 211: instance -> provider_name

CLAUDE.md
```diff
@@ -103,9 +103,8 @@ Full reference: `.env.template` and `config/config.yaml`
   - `PORT` (8080)
   - `GOMODEL_MASTER_KEY` (empty = unsafe mode)
   - `BODY_SIZE_LIMIT` ("10M")
-  - `ENABLE_PASSTHROUGH_ROUTES` (true: Enable provider-native passthrough routes under /p/{provider}/...)
-  - `ALLOW_PASSTHROUGH_V1_ALIAS` (true: Allow /p/{provider}/v1/... aliases while keeping /p/{provider}/... canonical)
-  - `ENABLED_PASSTHROUGH_PROVIDERS` (openai,anthropic,openrouter,zai,vllm: Comma-separated list of enabled passthrough providers)
+  - `ENABLE_PASSTHROUGH_ROUTES` (true: enable `/p/{instance}/...`; `instance` is the YAML key under `providers:`, not the provider type)
```
SantiagoDePolonia, line 106: instance -> provider_name

.env.template
```diff
@@ -15,14 +15,15 @@
...
 # ENABLE_PASSTHROUGH_ROUTES=true
 
-# Allow optional /p/{provider}/v1/... passthrough aliases while keeping /p/{provider}/... canonical (default: true)
+# Allow optional /p/{instance}/v1/... passthrough aliases while keeping /p/{instance}/... canonical (default: true)
 # ALLOW_PASSTHROUGH_V1_ALIAS=true
 
-# Comma-separated list of provider types enabled for /p/{provider}/... passthrough (default: openai,anthropic,openrouter,zai,vllm)
-# ENABLED_PASSTHROUGH_PROVIDERS=openai,anthropic,openrouter,zai,vllm
+# Per-instance passthrough opt-out: set passthrough_disabled: true on a provider in config/config.yaml
+# (there is no env allowlist; all configured instances with passthrough support are eligible by default)
```
SantiagoDePolonia, line 26: This line - should be removed "# (there is no env allowlist; all configured instances with passthrough support are eligible by default)"

internal/core/user_path.go
```diff
@@ -6,7 +6,14 @@ import (
 	"strings"
 )
 
-const UserPathHeader = "X-GoModel-User-Path"
+const (
+	// RequestIDHeader is the HTTP header GoModel uses to correlate requests
+	// across the gateway and its upstream calls. Clients may supply their own
+	// value; if absent a UUID is generated and stamped on every response.
+	RequestIDHeader = "X-GoModel-Request-ID"
```
SantiagoDePolonia, line 13: I guess it should not be at the `./user_path.go` file?

SantiagoDePolonia (commented):
I've added some minor comments during review, but this comment is the most important.

Generally I think that we shouldn't merge this and I think that it's my bad that I didn't communicate the requirements clearly.

What's well done:
- We should use provider_name instead of provider_type(named by you as instance) for sure
- This layer is more transparent then the previous solution
     - Errors should be returned in the original provider shape

---

The requirements are that we should be able to use it with guardrails and track the usage and have audit_logs for those requests. Also, we should be able to map some of the requests to make them understandable for our app at WhiteBoxPrompt. It unlocks as with enabling guardrails for most of the cases where GoModel can understand the Semantic of the request (what's the message, what model was used, etc) 

---
Summary generated with AI:

  1. High: passthrough usage logging regresses in the branch. internal/server/handlers.go:187 now builds a
     passthrough service that only has a response handler, and /internal/server/passthrough_service.go:12 only
     forwards bytes. In main, the old passthrough path still attached SSE usage/
     audit observers at internal/server/passthrough_support.go:283. So this PR
     is cleaner, but it misses the explicit goal of supporting passthrough
     usage, and it downgrades streaming observability.
  2. High: the default passthrough guardrail wiring is effectively a no-op. 
     /internal/server/http.go:265 skips
     workflow resolution for /p/..., but 
     internal/app/app.go:305 passes a workflow-based guardrail patcher, and that
     patcher resolves its pipeline from request context at internal/guardrails/.  workflow_executor.go:35. With
     no workflow in context, passthrough requests do not get request-scoped
     guardrails or other workflow-based mechanisms in the normal app wiring.
  3. High: even if a custom passthrough patcher is injected, request mutation is
     explicitly disabled. /internal/
     server/passthrough_middleware.go:86 states that guardrail modifications are
     intentionally discarded, and the handler still forwards the original body.
     That conflicts with the stated goal that content/header modifications
     should be possible.
  4. High: passthrough middleware fully buffers opaque bodies, twice. /internal/server/
     passthrough_middleware.go:36 reads the whole body for best-effort model
     extraction, and /internal/server/
     passthrough_middleware.go:96 reads it again for guardrail precheck. That is
     the wrong tradeoff for multipart/files/binary passthrough and works against
     the “all providers” target.

PR #255, opened 2026-04-22
fix(responsestore): use read lock for MemoryStore.Get on the happy path

hobostay:
## Summary
- `MemoryStore.Get()` was acquiring an exclusive write lock (`sync.Mutex.Lock()`) for every read operation, serializing all concurrent lookups
- The write lock was only needed for expired-entry cleanup and deletion, which are rare compared to normal reads
- Restructured `Get` to acquire `RLock` for the common case (entry found, not expired) and only upgrade to a write lock when an expired entry needs to be removed

## Test plan
- [x] Existing tests pass (`go test ./internal/responsestore/...`)
- [x] Verified compilation with `go build ./internal/responsestore/...`

## Summary by CodeRabbit

* **Refactor**
  * Enhanced memory store concurrency handling for improved performance under concurrent access patterns.

### DIFF
--- a/internal/responsestore/store_memory.go
+++ b/internal/responsestore/store_memory.go
@@ -104,19 +104,22 @@ func (s *MemoryStore) Create(_ context.Context, response *StoredResponse) error 
 // Get retrieves one response snapshot by id.
 func (s *MemoryStore) Get(_ context.Context, id string) (*StoredResponse, error) {
 	now := time.Now().UTC()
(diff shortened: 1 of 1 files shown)


SantiagoDePolonia:
@hobostay There are the serious issues with the PR described above. If you can solve them I think I can marge it.

SantiagoDePolonia:
stale

PR #256, opened 2026-04-22
fix(server): flush non-SSE passthrough responses after copy

hobostay:
## Summary
- `proxyPassthroughResponse` correctly flushes SSE streams after each chunk but does not flush the response writer for non-SSE responses
- When middleware (e.g., response capture) wraps the ResponseWriter with buffering, the final data may be held in the buffer and never delivered to the client
- Added an explicit `Flush` after `io.Copy` for non-SSE passthrough responses, matching the pattern already used for the SSE path

## Test plan
- [x] Existing tests pass (`go test ./internal/server/...`)
- [x] Verified compilation with `go build ./internal/server/...`

### DIFF
--- a/internal/server/passthrough_support.go
+++ b/internal/server/passthrough_support.go
@@ -305,5 +305,8 @@ func (s *passthroughService) proxyPassthroughResponse(c *echo.Context, providerT
 	if _, err := io.Copy(c.Response(), resp.Body); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


SantiagoDePolonia:
Hi @hobostay,
Thank you for the effort you put into hardening GoModel! I'm merging this, and possibly your other PRs as well.

PR #257, opened 2026-04-22
fix(responsecache): propagate write errors in writeCachedResponse

hobostay:
## Summary
- `writeCachedResponse` discarded response write errors with `_, _ = c.Response().Write(cached)`
- If the client disconnects during cache replay (e.g., connection reset), the function returned `nil`, causing the semantic cache to report a successful hit while the client received nothing
- Propagated write errors so the caller can log the failure and the upstream handler can react appropriately

## Test plan
- [x] Existing tests pass (`go test ./internal/responsecache/...`)
- [x] Verified compilation with `go build ./internal/responsecache/...`


SantiagoDePolonia:
@hobostay Hey again and thank you!
If you want me to merge it, please fix the issue described in the comment at the "Comments Outside Diff (1)" section.

SantiagoDePolonia:
Not active

PR #264, opened 2026-04-22
feat(providers): per-model metadata overrides for custom endpoints

epicfilemcnulty:
## Summary
                                                                                                                                                              
Lets operators declare per-model metadata (context window, pricing, capabilities, …) directly in `config.yaml` so `/v1/models` advertises accurate values for local models (Ollama) and other custom endpoints whose IDs are not in the upstream `ai-model-list` registry.
                                                                                                                                                              
`providers.<name>.models` now accepts **either** a plain model ID (as before) **or** a `{id, metadata}` object. When `metadata` is supplied it is merged onto the remote-registry entry during registry enrichment; operator-supplied values win per-field.
                                                                                                                                                              
## Motivation                                                                                                                                             

Right now, if you register a local model under Ollama (or any custom OpenAI-compatible endpoint), the gateway has no way to know its real context window, output limit, pricing, or capability set. Clients calling `/v1/models` get an unenriched entry with generic defaults.
                                                                                                                                                            
This PR gives operators a single place to declare that metadata without patching the registry or maintaining a fork of the upstream model list.             
   
## What's new                                                                                                                                               
                                                        
Extended YAML shape (backwards compatible — strings still work):                                                                                            
                                                        
```yaml                                                                                                                                                     
providers:                                            
  nippur:
    type: "ollama"
    base_url: "http://127.0.0.1:8080/v1"
    models:
      - id: GLM-4.7-Flash                                                                                                                                   
        metadata:
          display_name: "GLM 4.7 Flash (local)"                                                                                                             
          context_window: 131072                     
          max_output_tokens: 8192
          modes: ["chat"]                                                                                                                                   
          capabilities:
            tools: true                                                                                                                                     
          pricing:                                    
            currency: USD
            input_per_mtok: 0
      - Gemma4-31B   # plain string still supported
```                                                
                                                                                                                                                              
Supported metadata fields: `display_name`, `context_window`, `max_output_tokens`, `modes`, `capabilities`, `pricing`.
                                                                                                                                                              
## Behavior & scope                                   
                                                                                                                                                              
- Merge is per-field, not whole-object — operator values win for fields they set; unset fields fall back to the registry entry (or stay empty if the model
  is not in the registry).                                                                                                                                    
- Env-var-sourced model lists (`*_MODELS`) remain plain ID lists and never carry metadata.
- No change to `ProviderConfig.Models` (still `[]string`) — metadata travels through a parallel `ModelMetadataOverrides map[string]*core.ModelMetadata`     
  field keyed by raw model ID.                                                                                                                                
- `RawProviderConfig.Models` is the one type change: `[]string` → `[]RawProviderModel`. Internal only; YAML remains backwards compatible via a custom       
  unmarshaler in `config/provider_models.go`.                                                                                                                 
                                                                                                                                                              
## Tests                                              
                                                                                                                                                              
- New unit tests:                                     
- `config/provider_models_test.go` — unmarshal shapes (string, object, mixed), accessor helpers.                                                          
- `internal/modeldata/merge_test.go` — field-level merge precedence.                            
- `internal/providers/registry_metadata_override_test.go` — end-to-end override against the enriched registry.                                            
- Existing provider-env unit tests updated to the new `RawProviderModel` element type.                                                                      
- `go build ./... && go test ./...` clean on the branch.                                                                                                    
                                                                                                                                                              
## Observability                                                                                                                                            
                                                                                                                                                              
Registry log line already distinguishes `metadata_enriched` / `metadata_total` / `metadata_providers`, so operators can see overrides being applied at startup without extra wiring.                                                                                                                               
                               
## Checklist                                                                                                                                                
                                                        
- [x] Unit tests cover the new config shape, the merge, and the override path.
- [x] Backwards compatible — plain-string `models:` entries behave exactly as before.                                                                       
- [x] `config/config.example.yaml` has a commented-out real-world example.           
- [x] `CLAUDE.md` documents the new config knob.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Provider configs may list models as IDs or per-model objects; operator-supplied metadata (display name, context window, modes, capabilities, pricing) merges onto model info with operator fields taking precedence; ordering preserved and empty entries ignored. Overrides apply even when remote model lists are absent.

* **Documentation**
  * Expanded docs and example config demonstrating per-model metadata and merge behavior.

* **Tests**
  * Added comprehensive tests for YAML parsing, merge semantics, deep-cloning, and registry override behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


SantiagoDePolonia:
@epicfilemcnulty Thank you for your effort and for signaling what is needed!

It actually was on the roadmap, described as "Editable model pricing for accurate cost tracking and budgeting": https://gomodel.enterpilot.io/docs/about/roadmap

Generally, I plan to add a convenient way to override pricing-related metadata from the UI/dashboard (maybe other metadata as well?). But this doesn't seem to conflict much with this override mechanism at this point. So, it's a good improvement!

If you can fix those issues reported by the AI agents, I think I can safely merge it. (One comment is hidden under "Comments Outside Diff (1)".)

Thank you again for you contribution Vladimir!

epicfilemcnulty:
@SantiagoDePolonia thanks for looking through it. And thanks for the GoModel, btw, finally I was able to replace LiteLLM :)

To be frank, I'm not that concerned about pricing metadata (although other users might be), but IMO the accurate context window info is essential from the client POV -- it helps the harness to make desicions about context compaction & visualise context usage.

Let me fix all the issues raised by the agents then...

SantiagoDePolonia:
@epicfilemcnulty You're welcome! Good to hear that you replaced LiteLLM with GoModel!

BTW, could you describe exactly how you use GoModel? What do you plug in downstream? What problems does it solve for you? I'm curious how people are using it. Feel free to join our Discord if you'd rather not describe it here publicly.

epicfilemcnulty:
@SantiagoDePolonia 

There is nothing secret about my use case, but it is pretty simple. 

I don't really use (at least now) the advanced GoModel features like multi-tenancy and the like...I'm hosting local models on my home server via llama-server, and use Openrouter API for remote models. 

And I'm working on my minimal coding agent harness. Obviously it's easier to just implement one API on the client side, so I was using LiteLLM to route requests between local llama-server/openrouter API, and the harness was "LiteLLM" focused, now it's GoModel-centric :)

SantiagoDePolonia:
"minimal coding agent harness" - I'm curious what's behind this. Is it like your own piece of code? Or OpenCode?

epicfilemcnulty:
@SantiagoDePolonia nothing to write home about, pretty much the same thing is Pi, but in Lua (see [here](https://codeberg.org/latimar/lilpacks/src/branch/master/agent) if you are interested, but it's a rabbit hole -- it's designed to work as a mode in my own shell)

SantiagoDePolonia:
Thank you for linking GoModel there! It helps!

SantiagoDePolonia:
From my perspective, it's ready to be merged. The latest AI generated is nothing serious. It's up to you whether you want to fix it. I will merge it tomorrow.

epicfilemcnulty:
Yep, I think it's ready, and the last AI review is not quite right -- there is no immediate warning about a misspelled model, right, but there is a way to detect this -- hit the `/v1/models` endpoint and you will see that metadata is missing. I don't think it's worth adding another check just for this.

PR #266, opened 2026-04-23
fix(providers): fall back to configured models when upstream /models fails

mongrelion:
Problem triggered by configuring `type: openai` with a custom base URL for an API (e.g. OpenCode Go) that does not expose an OpenAI-compatible `/models` endpoint.

GoModel received HTML instead of JSON on discovery, failed to parse it, marked the provider "unhealthy", and discarded the explicitly configured model list entirely — even though those models work perfectly fine for chat completions. Oracle already had this fallback (PR #246), but every other OpenAI-compatible provider silently dropped configured models on discovery failure. This change makes them behave consistently with Oracle and respects the user's explicit model configuration as a reliability guarantee.

Value: providers that users have explicitly configured with a known list of working models no longer go "unhealthy" just because their upstream doesn't support the /models endpoint or returns an unexpected response shape. The fallback preserves any metadata returned by upstream when it is available, and degrades gracefully to bare model IDs otherwise. 

---

* refactor(openai-compatible): extract doListModels from ListModels so the new fallback logic reuses the same upstream call path; add a helper that deduplicates and trims configured model names for correctness.

* fix(openai-compatible): make CompatibleProvider.ListModels fall back to explicitly configured models when the upstream /models endpoint returns an error (including non-JSON responses such as HTML pages), so providers with pre-defined model lists stay usable even when discovery fails.

* fix(openai, openrouter, vllm): pass opts.Models through to CompatibleProviderConfig so these provider types benefit from the new fallback logic; Azure gets its own fallback in its custom ListModels method since it calls a different endpoint (/openai/models).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed model listing failures across Azure, OpenAI, OpenRouter, and vLLM providers. When the upstream model endpoint is unavailable or returns an error, the system now gracefully falls back to your configured models instead of failing completely. Non-critical issues are logged as warnings.

### DIFF
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -2,6 +2,7 @@ package azure
 
 import (
 	"context"
+	"log/slog"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -30,6 +31,7 @@ type Provider struct {
 	resourceProvider       *openai.CompatibleProvider
 	openAIResourceProvider *openai.CompatibleProvider
 	apiVersion             string
+	configuredModels       []string
 }
 
 func New(providerCfg providers.ProviderConfig, opts providers.ProviderOptions) core.Provider {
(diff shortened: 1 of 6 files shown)


SantiagoDePolonia:
@mongrelion Hey 👋
Thank you for your patience and your initiative. I'm merging this BUT I'm going to refactor it deeply in the upcoming PR and add:
- support for env variables - in GoModle (almost) everything should be configurable with env variables

Feel free to suggest any changes in the future and to let me know about any issues.

SantiagoDePolonia:
I've refactored it here: https://github.com/ENTERPILOT/GoModel/pull/268

mongrelion:
@SantiagoDePolonia Thank you for looking into this. I was overwhelmed by the amount of feedback that I got from the reviewer bot. I'm happy that the feature got it, even better that it got refactored into something that fits better the style of the rest of the project.

Cheers :)

SantiagoDePolonia:
Happy to help!

Those bots are useful for detecting potential issues and improving SEO. :)

Feel free to suggest changes or report any issues in the future. I'm always happy to hear what GoModel users need or struggle with.

PR #298, opened 2026-05-03
fix(passthrough): preserve original provider name in route resolution

vfeitoza:
The semantic enrichment middleware was overwriting info.Provider with the
  resolved type (e.g. 'anthropic') before caching, causing passthroughExecutionTarget
  to route by type instead of the original configured name (e.g. 'teste').

  Added info.ProviderName field to preserve the original route name through the
  resolution pipeline. The execution target now prefers the cached provider name
  when available, ensuring requests route to the correct configured instance.

## Summary by CodeRabbit

* **New Features**
  * Added support for name-based provider routing, allowing you to specify a particular configured provider instance by name for pass-through requests instead of relying solely on provider type-based resolution.

* **Improvements**
  * Enhanced debug logging for provider resolution to aid in troubleshooting routing behavior.

### DIFF
--- a/internal/core/passthrough.go
+++ b/internal/core/passthrough.go
@@ -8,10 +8,11 @@ import (
 
 // PassthroughRequest is the transport-oriented request for opaque provider-native forwarding.
 type PassthroughRequest struct {
-	Method   string
-	Endpoint string
-	Body     io.ReadCloser
-	Headers  http.Header
+	Method       string
+	Endpoint     string
+	Body         io.ReadCloser
+	Headers      http.Header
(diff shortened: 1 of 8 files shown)


SantiagoDePolonia:
Thank you!

LGTM, merging.

I should mention that the current passthrough API is in Beta and it might change in the upcoming days. From the user's perspective, it should work the same, but be more reliable and support more AI model providers. The current implementation only supports Anthropic and OpenAI providers explicitly.

Feel free to report any suggestion, issue or submit a PR in the future.

PR #311, opened 2026-05-09
chore(swagger): reduce dependencies

SantiagoDePolonia:
## Summary
- Serve Swagger UI directly from `swaggo/files/v2` assets and `swaggo/swag/v2` specs
- Remove the `echo-swagger` and `swaggo` v1 dependency path
- Add coverage for `doc.yaml` and static Swagger UI assets

## Tests
- `go test ./...`
- commit hooks: `make test-race`, `go mod tidy`, hot-path performance guard, `make lint`
- `mint validate` skipped because local Node is 25.9.0 and Mintlify requires an LTS Node version

## Release Notes

* **Documentation**
  * Updated development guide for API specification tooling configuration.

* **Refactor**
  * Refactored API documentation serving mechanism.
  * Updated core project dependencies.

* **Tests**
  * Added test coverage for API documentation endpoints (YAML format and static assets).

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -44,10 +44,9 @@ Use `make swagger` to regenerate the checked-in Swagger docs package and
 `docs/openapi.json`.
 
 GoModel intentionally uses `github.com/swaggo/swag/v2/cmd/swag`. The Swagger UI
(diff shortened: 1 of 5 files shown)


SantiagoDePolonia:
Not needed - it doesn't affect the build size

PR #317, opened 2026-05-10
fix(auditlog): reconstruct streamed tool calls

SantiagoDePolonia:
## Summary
- reconstruct streamed chat audit responses per choice instead of content-only
- accumulate streamed `delta.tool_calls` id/type/name and concatenated arguments
- preserve provider metadata and trailing usage in captured streaming response bodies

## Verification
- `go test ./internal/auditlog ./internal/streaming ./internal/responsecache`
- `go test ./...`
- commit hook: `make test-race`, `go mod tidy`, hot-path performance guard, `make lint`
- live Bedrock repro with `amazon.nova-micro-v1:0`: audit row stored `tool_calls[0]` as `get_weather` with `{"city":"Paris"}`, `finish_reason=tool_calls`, and usage total tokens

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for streamed chat responses: text-only, tool-call-only, interleaved text/tool streams, parallel tool-call aggregation, sparse-index fallbacks, skipping incomplete tool calls, trailing usage capture, and truncation behavior when budget is exhausted.

* **Improvements**
  * Streaming response builder now emits richer multi-choice chat outputs with preserved metadata, deterministic tool-call ordering, robust handling of streamed tool-call arguments, and improved output-text aggregation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ENTERPILOT/GoModel/pull/317)

### DIFF
--- a/internal/auditlog/auditlog_test.go
+++ b/internal/auditlog/auditlog_test.go
@@ -1159,6 +1159,23 @@ func TestStreamResponseBuilderChatCapturesTrailingUsageChunk(t *testing.T) {
 	}
 }
 
+func TestAppendLimitedStreamTextMarksTruncatedWhenBudgetAlreadyFull(t *testing.T) {
+	builder := &streamResponseBuilder{contentLen: MaxContentCapture}
+	var dst strings.Builder
+
+	appendLimitedStreamText(builder, &dst, "x")
+
+	if !builder.truncated {
+		t.Fatal("truncated = false, want true when a non-empty chunk arrives after the capture budget is full")
+	}
+	if got := dst.String(); got != "" {
+		t.Fatalf("captured text = %q, want empty", got)
+	}
+	if got := builder.contentLen; got != MaxContentCapture {
+		t.Fatalf("contentLen = %d, want %d", got, MaxContentCapture)
+	}
+}
+
 func buildChatStreamResponseForTest(t *testing.T, events ...string) map[string]any {
 	t.Helper()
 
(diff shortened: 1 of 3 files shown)


SantiagoDePolonia:
Addressed the review follow-ups in `e8a0a65`:

- Sparse fallback indexes now use `max(existing)+1` for both choices and tool calls, so index-less chunks do not collide with sparse provider indexes.
- Tool-call states that never receive a `function` delta are no longer serialized, while id/type from earlier deltas is still preserved when function data arrives later.
- Added regression coverage for sparse choice indexes, sparse tool-call indexes, and id/type-only orphan tool-call deltas.

I left the shared `MaxContentCapture` budget as-is intentionally: audit streaming capture has a single global body safety cap and already marks `ResponseBodyTooBigToHandle` when it truncates. Splitting the cap per field would allow unbounded total growth across many choices/tool calls and change the existing memory-safety behavior.

Validation rerun:
- `go test ./internal/auditlog`
- `go test ./internal/streaming ./internal/responsecache`
- `go test ./...`
- commit hook: `make test-race`, `go mod tidy`, hot-path performance guard, `make lint`

SantiagoDePolonia:
Addressed the latest review findings in `7bd1f9a`:

- `appendLimitedStreamText` now marks `ResponseBodyTooBigToHandle` when a non-empty chunk arrives after the stream capture budget is already exactly full, and still marks truncation when a chunk is partially captured.
- Added a short comment documenting the `message["content"] = nil` vs `""` distinction for tool-only vs truly empty assistant messages.
- Added regression coverage for the exact-budget overflow case.

Validation rerun:
- `go test ./internal/auditlog`
- `go test ./internal/streaming ./internal/responsecache`
- `go test ./...`
- commit hook: `make test-race`, `go mod tidy`, hot-path performance guard, `make lint`

PR #332, opened 2026-05-15
fix(passthrough): track usage for non-streaming responses and extract…

vfeitoza:
… model from opaque bodies

Non-streaming passthrough responses were not logging usage data because the response body was copied directly to the client without parsing. Additionally, when the request body was not fully captured (chunked transfer), the model field was not extracted for opaque passthrough routes because the peek logic required both model and provider to apply selector hints.

- Read non-streaming response body and extract usage via ExtractFromCachedResponseBody before forwarding to the client
- Apply body selector hints for BodyModeOpaque when only model is found, ensuring passthrough routes populate PassthroughRouteInfo.Model even without a provider field in the request body

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved usage tracking and logging for non-streaming provider responses, including accurate model/provider resolution and pricing calculation.
  * Enhanced request body parsing with more robust selector hint application for increased reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ENTERPILOT/GoModel/pull/332)

### DIFF
--- a/internal/server/passthrough_support.go
+++ b/internal/server/passthrough_support.go
@@ -311,12 +311,55 @@ func (s *passthroughService) proxyPassthroughResponse(c *echo.Context, providerT
 		return nil
 	}
 
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return handleError(c, core.NewProviderError(providerType, http.StatusBadGateway, "failed to read provider passthrough response body", err))
+	}
+
+	workflow := core.GetWorkflow(c.Request().Context())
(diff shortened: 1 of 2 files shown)


SantiagoDePolonia:
The main issue was fixed in GoModel many days ago. Usage tracking for passthrough should work well. Although the whole passthrough might be rebuild soon

The fix for second issue is coming here: github.com/ENTERPILOT/GoModel/pull/795

Closing this one because it's stale. Thank you for reporting!

If you want to work on anything - please add it as an issue OR add a comment to existing issues that you will work on this. Your contributions are welcome!

PR #333, opened 2026-05-15
feat(models): add configurable model ID format for GET /v1/models

vfeitoza:
Problem:

The GET /v1/models endpoint returns model IDs in qualified format (provider/model, e.g. "anthropic/claude-sonnet-4-6"). OpenAI-compatible clients such as Claude CLI, Cursor, and other tools expect simple model IDs (e.g. "claude-sonnet-4-6"). When these clients fetch the model list to validate a requested model against available models, the qualified IDs never match the simple ID the client intends to send, resulting in "model not found" errors on the client side.

The internal routing already resolves both qualified and unqualified selectors correctly — the problem is exclusively in the public listing response format.

Solution:

Add a new configuration option `models_endpoint_id_format` (env: MODELS_ENDPOINT_ID_FORMAT) that controls the model ID format returned by GET /v1/models. Supported values:

- "qualified" (default): provider/model format (current behavior, no breaking change)
- "unqualified": simple model ID only, with OwnedBy set to the provider name for disambiguation
- "both": returns both qualified and unqualified entries, allowing clients that understand either format to work seamlessly

Implementation:

- config/models.go: ModelsEndpointIDFormat type with validation and normalization helpers following existing ConfiguredProviderModelsMode pattern
- config/config.go: default value set to "qualified"
- internal/providers/registry.go: ListModelsWithFormat() dispatches to ListPublicModels (qualified), listModelsUnqualified, or listModelsBoth. Unqualified deduplicates by model ID (first provider wins, matching routing behavior). Both merges qualified + unqualified with ID-based deduplication.
- internal/providers/router.go: Router gains modelsEndpointIDFormat field and SetModelsEndpointIDFormat setter. ListModels uses the configured format via interface assertion on the lookup.
- internal/app/app.go: wires config to Router after provider init
- config/config.example.yaml, .env.template: documents the new option
- internal/providers/router_test.go: TestRouterListModelsWithFormat covers all three format modes

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable formatting for model IDs returned by the models endpoint. Users can now choose between "qualified" (including provider prefix), "unqualified" (model name only), or "both" formats, configurable via the `MODELS_ENDPOINT_ID_FORMAT` environment variable or configuration file.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ENTERPILOT/GoModel/pull/333)

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -114,6 +114,8 @@
 # fallback (default): use configured models only when upstream /models fails, is nil, or is empty.
 # allowlist: expose only the configured models for providers that define a list, and skip their upstream /models calls.
 # CONFIGURED_PROVIDER_MODELS_MODE=fallback
+# Model ID format for GET /v1/models: "qualified" (provider/model), "unqualified" (model only), or "both"
+# MODELS_ENDPOINT_ID_FORMAT=qualified
 # Examples: OPENROUTER_MODELS=..., OPENROUTER_EU_MODELS=..., AZURE_MODELS=..., VLLM_MODELS=...
 
 # Fallback & Workflow Configuration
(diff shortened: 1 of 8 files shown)


SantiagoDePolonia:
The idea here is great! {provider}/ prefix should be skippable 

I've started to work on this from the latest main here: https://github.com/ENTERPILOT/GoModel/pull/796

Soon it will land in GoModel release.

Thank you for the feedback and sorry for missing this one.

SantiagoDePolonia:
Closing, because it's stale and GoModel changed a lot since the PR was created

PR #363, opened 2026-06-01
test(minimax): use MiniMax-M3 in test fixtures

octo-patch:
## Summary
Update the MiniMax (minimax) provider test fixtures to reference `MiniMax-M3`, the current flagship model, in place of `MiniMax-M2.7`.

## Changes
- `internal/providers/minimax/minimax_test.go`: replace `MiniMax-M2.7` with `MiniMax-M3` in mock server responses, request payloads, and assertions.

## Why
The MiniMax provider is a passthrough to the upstream API and dynamically supports any model the upstream exposes via `/models`. The provider does not maintain a hardcoded model list or default, so no production code needs to change to support the new model. Aligning the test fixtures with the current flagship keeps the test suite representative of typical usage.

MiniMax-M3 is the new flagship model with 512K context window, 128K max output, and image input support, available across both OpenAI-compatible and Anthropic-compatible interfaces.

## Testing
- `go test ./internal/providers/minimax/...` passes
- `go vet ./internal/providers/minimax/...` clean

## Summary by CodeRabbit

* **Tests**
  * Updated minimax provider tests to validate compatibility with the latest model version.

### DIFF
--- a/internal/providers/minimax/minimax_test.go
+++ b/internal/providers/minimax/minimax_test.go
@@ -23,7 +23,7 @@ func TestChatCompletion_UsesBearerAuthAndChatEndpoint(t *testing.T) {
 		_, _ = w.Write([]byte(`{
 			"id":"chatcmpl-minimax",
 			"created":1677652288,
-			"model":"MiniMax-M2.7",
+			"model":"MiniMax-M3",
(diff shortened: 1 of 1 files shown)


SantiagoDePolonia (approved):
LGTM!

PR #372, opened 2026-06-05
feat(usage): track and cost STT/TTS interactions

SantiagoDePolonia:
## What & why

Audio endpoints (`/v1/audio/speech` and `/v1/audio/transcriptions`) were not recorded in usage tracking. Chat, responses, embeddings, and batches all log usage; audio was the gap. This adds usage + cost tracking for both.

Audio providers return opaque bytes with no token usage, so usage is derived locally at the audio service layer — providers and the router are untouched; the audio service is the single integration point, mirroring how the inference orchestrator logs usage.

## How usage is captured

- **Speech (TTS):** record input character count in `raw_data.input_characters`.
- **Transcription (STT):** parse the optional `usage` object from the proxied response — token-based (`gpt-4o-transcribe`) or duration-based (`whisper`, recorded as `raw_data.audio_seconds`). An entry is **always** written so the interaction stays observable even when the provider reports no usage (whisper without usage, or `text`/`srt`/`vtt` response formats).

## Cost

Wires the non-token pricing units already present in registry pricing but previously unused by the cost calculator:

- `per_character_input` × input characters (TTS)
- `per_second_input` × input audio duration (STT)

alongside the existing token-based path.

| model | priced by | result |
|---|---|---|
| `tts-1` / `tts-1-hd` | per-character | real cost |
| `whisper-1` | per-second | real cost |
| `gpt-4o-transcribe` | tokens | real cost (existing path) |
| `gpt-4o-mini-tts` | output audio duration | **uncosted** (see below) |

### Known limitation

Models priced by **output** audio duration (e.g. `gpt-4o-mini-tts` via `per_second_output`) remain uncosted: the gateway proxies opaque audio bytes and does not measure their duration. Documented in `applyAudioUnitCosts`. The usage interaction is still logged (with `input_characters`), just with `$0` cost.

## Verification

Verified live against OpenAI:

| model | usage | cost |
|---|---|---|
| `tts-1` | 50 chars | \$0.00075 |
| `whisper-1` | 4 sec | \$0.0004 |
| `gpt-4o-transcribe` | 42 tok | \$0.000195 |
| `gpt-4o-mini-tts` | 11 chars | \$0.00 (known gap) |

## Tests

- New `internal/usage/audio_test.go`: speech/transcription extraction (token, duration, no-usage, per-character cost, per-second cost, no-rate cases).
- New cases in `internal/server/audio_service_test.go`: usage is logged for speech & transcription, and not logged when usage is disabled.
- `go build`, `go vet`, full `go test ./...`, and pre-commit `make test-race` / `make lint` all pass.

## Summary by CodeRabbit

* **New Features**
  * Per-call audio usage tracking: records input characters for text-to-speech and audio duration for speech-to-text, with optional pricing applied.

* **Behavior**
  * Provider identity is consistently associated with usage entries.
  * Service returns a 502 when an upstream audio provider returns an empty response; no usage is recorded in that case.
  * Usage tracking can be disabled to suppress logging.

* **Tests**
  * Added comprehensive tests for audio usage extraction, pricing, logging, and nil-response handling.

### DIFF
--- a/internal/server/audio_service.go
+++ b/internal/server/audio_service.go
@@ -74,8 +74,11 @@ func (s *audioService) CreateSpeech(c *echo.Context) error {
 	if err != nil {
 		return handleError(c, err)
 	}
+	if resp == nil {
+		return s.respondAudio(c, resp) // emits the 502 guard; no usage for a failed call
+	}
 	s.logUsage(ctx, route, func(pricing *core.ModelPricing) *usage.UsageEntry {
-		return usage.ExtractFromSpeechRequest(req, route.requestID, route.providerType, pricing)
+		return usage.ExtractFromSpeechRequest(req.Input, route.requestID, route.model, route.providerType, pricing)
 	})
 	return s.respondAudio(c, resp)
 }
@@ -107,6 +110,9 @@ func (s *audioService) CreateTranscription(c *echo.Context) error {
 	resp, err := router.CreateTranscription(ctx, req)
 	if err != nil {
 		return handleError(c, err)
+	}
+	if resp == nil {
+		return s.respondAudio(c, resp) // emits the 502 guard before resp.Data is read
 	}
 	s.logUsage(ctx, route, func(pricing *core.ModelPricing) *usage.UsageEntry {
 		return usage.ExtractFromTranscriptionResponse(resp.Data, route.requestID, route.model, route.providerType, pricing)
--- a/internal/server/audio_service.go
+++ b/internal/server/audio_service.go
@@ -11,6 +11,7 @@ import (
 
 	"gomodel/internal/auditlog"
 	"gomodel/internal/core"
+	"gomodel/internal/usage"
 )
 
 // audioService adapts Echo requests to the model-routed audio provider for the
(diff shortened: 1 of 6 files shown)


SantiagoDePolonia:
Addressed the review feedback in a2a9918:

**✅ Fixed — nil-response guard** (greptile P2 + CodeRabbit 🔴 on `audio_service.go`)
Both handlers now short-circuit to the `respondAudio` 502 guard when the provider returns `(nil, nil)`, *before* `logUsage`. This removes the transcription panic on `resp.Data` and stops a usage row being written for a request that ends up 502. Added `TestAudio_NilResponseSkipsUsage` (speech + transcription) to lock it in.

**✅ Fixed — log resolved model for speech** (greptile P1)
Speech usage now records `route.model` (the resolved route model) instead of the raw `req.Model`, matching the transcription extractor and the pricing lookup. `ExtractFromSpeechRequest` now takes `(input, requestID, model, provider, …)`, symmetric with `ExtractFromTranscriptionResponse`.

**❌ Skipped — "resolve pricing with `providerType`"** (CodeRabbit 🟠 on `audio_service.go:200`)
This one doesn't hold up. The code intentionally mirrors the inference orchestrator's `effectivePricingProvider`, which resolves **provider name first, then type**. `ModelRegistry.ResolvePricing` looks up `modelsByProvider[selector]`, and that map is keyed **by provider instance name** (see `registry_init.go` / `registry_cache.go`), so name-first is the correct lookup — and it falls back to the type via `metadataProviderType` when needed. Flipping to type-first would diverge from the canonical chat/embeddings usage path. Left as-is.

All CI was already green; `go build`, `go vet`, full `go test ./...`, and pre-commit `make test-race` / `make lint` pass after the changes.

PR #374, opened 2026-06-07
fix: extract system messages from messages array in Anthropic API

vfeitoza:
The Anthropic Messages API (/v1/messages) only accepts "user" or "assistant" roles in the messages array. System prompts must be passed in the top-level "system" field.

Previously, messages with role "system" in the messages array were rejected with error: "messages[N].role must be \"user\" or \"assistant\""

This issue only occurred when using Claude Opus 4.8 model, as other models did not exhibit this behavior

This change extracts messages with role "system" from the messages array and prepends them to the system prompt, allowing clients to send system content in either format.

Fixes compatibility with Claude Code and other clients that include system messages in the messages array.

## Summary by CodeRabbit

* **New Features**
  * System messages can now be included within the messages array and will be automatically combined with any top-level system instructions.


SantiagoDePolonia:
Merged! Thank you for your PR! It's actually great feedback, because I see what people need, how they use GOModel and what I should be focued on.

PR #376, opened 2026-06-08
fix(usage): cost text-to-speech by output audio duration

SantiagoDePolonia:
## Summary

Text-to-speech models priced by **output audio duration** — `gpt-4o-mini-tts` and the rest of the gpt-4o TTS family — were logged with a **silent `total_cost` of 0** and no caveat. Their registry pricing carries `per_second_output` (e.g. `$0.00025/s` = $0.015/min) but **no** `per_character_input`, and the cost path only applied `PerCharacterInput` / `PerSecondInput`. Legacy per-character models (`tts-1`) and all STT models were already costed correctly.

This was found during release smoke testing: `gpt-4o-mini-tts` recorded `input_characters` but `total_cost=0`, while `tts-1` (per-character) and `whisper-1`/`gpt-4o-transcribe` (per-second / per-token) priced fine.

## Fix

The gateway already has the synthesized audio bytes at usage time, so it can measure the duration that `per_second_output` needs:

- **WAV** (`response_format=wav`): parsed from the self-describing RIFF/WAVE header (byte rate × data-chunk size), tolerant of extra chunks and unknown/streamed data sizes.
- **PCM** (`response_format=pcm`): OpenAI's headerless 16-bit / 24 kHz / mono stream → `len/48000` bytes-per-second.
- **Compressed** (mp3 — the API default — opus, aac, flac): cannot be measured without a decoder, so they now surface `costs_calculation_caveat: "audio output duration unmeasured for format: <fmt>"` instead of a silent `$0`.

`PerSecondOutput` is applied on the **output** side; the new `audio_output_seconds` / `audio_output_format` keys live in `raw_data`. **Transcription is untouched** — STT rows set `audio_seconds` (input) and never the new output keys, so per-second-output STT models like `whisper-1` are unaffected.

## User-visible impact

- TTS usage rows for output-duration-priced models now report real cost for uncompressed output and an explicit caveat (not a silent `$0`) for compressed output.
- New `raw_data` fields on `/v1/audio/speech` usage entries: `audio_output_seconds` (when measured) and `audio_output_format`.
- No public API change; no config change.

## Testing

- New unit tests: WAV/PCM/compressed duration measurement, unknown-data-size fallback, format normalization, and cost paths (WAV output → `PerSecondOutput`, PCM output, mp3 → caveat).
- New end-to-end service test: a WAV-returning `gpt-4o-mini-tts` is priced via `PerSecondOutput` through the full `CreateSpeech` path.
- `go test ./...` green; `make lint`, `make test-race`, and the hot-path perf guard pass.
- **Live verified** against the release e2e stack on SQLite, PostgreSQL, and MongoDB:
  - WAV → `audio_output_seconds` measured, `total_cost = secs × $0.00025` (e.g. 5s → `$0.00125`) on all three backends.
  - MP3 → `total_cost=0` **with** caveat `audio output duration unmeasured for format: mp3`.
  - PCM → measured via fixed rate.

## Follow-up (not in this PR)

Measuring duration for compressed formats (mp3/opus/aac/flac) would need a decoder/parser; left as a caveat for now. The text-token input cost for these models (`input_per_mtok`) is also not applied since the gateway records characters, not tokens — negligible next to the audio-output cost, and out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Usage billing for text‑to‑speech now derives from synthesized audio output (measured duration and normalized format) when available.

* **Improvements**
  * Improved audio format normalization and duration estimation; billing records output-audio-seconds and adds caveats when duration is unmeasurable.
  * Pricing now separates input vs. output audio costs; gateway respects upstream Content-Type for returned audio.

* **Tests**
  * Added tests for WAV/PCM duration measurement, format normalization, output-duration pricing, and upstream Content-Type handling.

### DIFF
--- a/internal/llmclient/client.go
+++ b/internal/llmclient/client.go
@@ -168,8 +168,11 @@ type Request struct {
 // Response represents an HTTP response
 type Response struct {
 	StatusCode int
-	Header     http.Header
-	Body       []byte
+	// ContentType is the upstream response Content-Type header, preserved so
+	// callers can describe the bytes actually returned (e.g. audio formats).
+	// Only this header is surfaced; the full upstream header set is not exposed.
+	ContentType string
+	Body        []byte
 }
 
 type requestScope struct {
@@ -616,9 +619,9 @@ func (c *Client) doRequest(ctx context.Context, req Request) (*Response, error) 
 	}
 
 	return &Response{
-		StatusCode: resp.StatusCode,
-		Header:     resp.Header,
-		Body:       body,
+		StatusCode:  resp.StatusCode,
+		ContentType: resp.Header.Get("Content-Type"),
+		Body:        body,
 	}, nil
 }
 
--- a/internal/llmclient/client.go
+++ b/internal/llmclient/client.go
@@ -168,6 +168,7 @@ type Request struct {
 // Response represents an HTTP response
 type Response struct {
 	StatusCode int
+	Header     http.Header
 	Body       []byte
 }
 
@@ -616,6 +617,7 @@ func (c *Client) doRequest(ctx context.Context, req Request) (*Response, error) 
 
 	return &Response{
 		StatusCode: resp.StatusCode,
+		Header:     resp.Header,
 		Body:       body,
 	}, nil
 }
--- a/internal/providers/openai/audio.go
+++ b/internal/providers/openai/audio.go
@@ -13,8 +13,10 @@ import (
 )
 
 // CreateSpeech implements OpenAI text-to-speech (POST /audio/speech). The upstream
-// returns binary audio, so the response is read raw and tagged with the content
-// type implied by the requested response_format.
+// returns binary audio; the response is read raw and tagged with the upstream
+// Content-Type so it describes the bytes actually returned (see
+// speechResponseContentType), which usage relies on to price output-duration
+// models.
 func (p *CompatibleProvider) CreateSpeech(ctx context.Context, req *core.AudioSpeechRequest) (*core.AudioResponse, error) {
 	if req == nil {
 		return nil, core.NewInvalidRequestError("audio speech request is required", nil)
@@ -35,9 +37,22 @@ func (p *CompatibleProvider) CreateSpeech(ctx context.Context, req *core.AudioSp
 		return nil, err
 	}
 	return &core.AudioResponse{
-		ContentType: core.SpeechResponseContentType(req.ResponseFormat),
+		ContentType: speechResponseContentType(raw, req.ResponseFormat),
 		Data:        raw.Body,
 	}, nil
+}
+
+// speechResponseContentType returns the authoritative media type of synthesized
+// speech. The upstream response Content-Type describes the bytes actually
+// returned, so it wins; it falls back to the type implied by the requested
+// response_format when the upstream omits the header.
+func speechResponseContentType(raw *llmclient.Response, format string) string {
+	if raw != nil {
+		if ct := strings.TrimSpace(raw.Header.Get("Content-Type")); ct != "" {
+			return ct
+		}
+	}
+	return core.SpeechResponseContentType(format)
 }
 
 // CreateTranscription implements OpenAI speech-to-text (POST /audio/transcriptions).
--- a/internal/providers/openai/audio.go
+++ b/internal/providers/openai/audio.go
@@ -48,7 +48,7 @@ func (p *CompatibleProvider) CreateSpeech(ctx context.Context, req *core.AudioSp
 // response_format when the upstream omits the header.
 func speechResponseContentType(raw *llmclient.Response, format string) string {
 	if raw != nil {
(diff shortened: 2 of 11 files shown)


SantiagoDePolonia:
Reviewed all bot findings, verified each against the code, and pushed fixes in 1a342d9.

**Fixed (4):**
- **Request format can misprice** (greptile, `audio_service.go`) — `speechResponseFormat` now trusts the response `Content-Type` over the requested `response_format`, since it describes the bytes actually returned. Verified live: OpenAI returns `audio/wav`, `audio/mpeg`, `audio/pcm`, `audio/ogg`, so wav/pcm stay measurable and a format mismatch falls through to the caveat path instead of mispricing.
- **Empty output loses caveat** (greptile, `audio.go`) — `audio_output_format` is now recorded regardless of body length; only the measured duration is gated on a non-empty body, so a compressed/empty response still surfaces the caveat.
- **Zero-length RIFF chunks treated as fatal** (coderabbit, `audio_duration.go`) — the walker no longer breaks on a zero-length chunk; it advances to the next header. (`size` is read from a `uint32`, so it is never negative — the old guard only ever fired on `size == 0`.) Added a regression test.
- **Fallback-branch tests** (coderabbit, `audio_service_test.go`) — `TestAudioSpeech_CostsOutputAudioDuration` is now table-driven, covering explicit `response_format`, `Content-Type` fallback, and the mp3 default.

**Skipped (2), no real-world trigger:**
- **RF64 WAVs** — RF64 is the >4 GB WAV variant; TTS synthesis output is a few MB at most and never emits it.
- **Malformed byteRate** — requires a corrupt WAV from the provider; OpenAI returns well-formed PCM WAV, and the RIFF/WAVE magic gate plus `byteRate == 0` guard already cover the realistic cases.

Full suite, `make lint`, `make test-race`, and the hot-path perf guard pass; re-verified live on the e2e stack (wav/pcm measured and priced; mp3/opus caveated at $0).

SantiagoDePolonia:
Addressed the conflicting-format suggestion in 9785107 — with a stronger variant.

The exact proposed row (`pcm` request + `audio/wav` Content-Type + WAV bytes) doesn't actually exercise the precedence logic: `measureSpeechDurationSeconds` detects the RIFF magic bytes first, so WAV data is measured the same whether the format string is `pcm` or `wav` — that row passes even if Content-Type precedence were removed.

The discriminating case is the mispricing direction: client requests a *measurable* format (`pcm`) but the provider actually returns mp3. I added:

```
{"content-type overrides requested format", "pcm", "audio/mpeg", mp3, 0, 0, true},
```

If the request format were (incorrectly) trusted, the 15-byte mp3 stub would be charged as `len/48000` PCM seconds with no caveat, failing the assertions. With Content-Type precedence it resolves to mp3 → unmeasured → caveat at $0, which is what the row asserts. So this genuinely locks in the fix.

SantiagoDePolonia:
Good catch on the 3/5 review — the content-type precedence was circular and I've fixed it at the root in d304054.

**The real problem:** `internal/providers/openai/audio.go` set the speech response type from `core.SpeechResponseContentType(req.ResponseFormat)`, i.e. derived from the request. So the usage layer "preferring `resp.ContentType` over the request" preferred two values that always agreed — no actual misprice protection, exactly as flagged.

**Fix:** preserve the real upstream `Content-Type` through the client:
- `llmclient.Response` now carries `Header http.Header` (populated in `doRequest`).
- `CreateSpeech` tags the response with the upstream `Content-Type`, falling back to the `response_format`-derived type only when the upstream omits it.

The response media type now describes the bytes actually returned, so the output-duration pricing path is authoritative and a transcoded/mismatched response is caveated instead of mispriced.

**Verified live against OpenAI:** `wav → audio/wav`, `pcm → audio/pcm` (both still measured/priced), `mp3 → audio/mpeg`, `opus → audio/opus` (caveated). Added provider tests covering upstream-honored, upstream-overrides-request, and fallback-to-request paths. Audio e2e S110–S114 pass; full suite, lint, test-race, and perf guard green.

SantiagoDePolonia:
Fixed the failing **Contract Replay Tests** and addressed the header nitpick in 574543a.

**CI failure** (`TestOpenAIReplayCreateSpeech`): the fixture returned `audio/mpeg` for *every* format while asserting a per-format type derived from the request — it encoded the pre-fix behavior. Since the provider now preserves the upstream `Content-Type`, I made the fixture realistic (`wav → audio/wav`, `opus → audio/opus`, matching what OpenAI actually returns) and assert the provider preserves it, plus a case where the upstream type overrides a divergent requested format. This is the test catching up to the corrected logic, not a regression.

**Header nitpick** (`llmclient.Response`): rather than expose the full `http.Header` (and risk a future caller logging sensitive upstream headers), I narrowed it to a single `ContentType string` populated from the upstream `Content-Type`. The only consumer needs exactly that, so it removes the footgun at the root and keeps the interface small.

Contract suite (`-tags=contract`), full suite, lint, test-race, and the perf guard all pass locally.

PR #386, opened 2026-06-10
fix: resolve gateway issues found by end-to-end testing

SantiagoDePolonia:
## Summary

A ~130-test end-to-end pass over the gateway (all providers, streaming + non-streaming, audio, all three storage backends, with DB-level verification) surfaced five bugs. This PR fixes all of them; each fix was re-verified live against real providers.

### 1. Conversations now work with `/v1/responses`
`POST /v1/conversations` stores conversations locally, but `/v1/responses` forwarded the `conversation` field verbatim upstream — so every gateway-created conversation ID 404ed at the provider. The gateway now resolves conversations locally: stored history is prepended to the input at prepare time, the field is stripped before dispatch, and the completed exchange is appended back to the store (streamed turns included, via a stream observer on `response.completed`).

**User-visible impact:** conversations become usable at all — including with chat-translated providers (Anthropic, Groq, …) and even across providers within one conversation. Unknown IDs return OpenAI's 404 contract; `conversation` + `previous_response_id` together is rejected like OpenAI does; conversation turns bypass the exact response cache (a hit would serve a stale turn and skip the history append).

### 2. mp3 TTS output is now costed
Duration-priced TTS models (`gpt-4o-mini-tts`) are billed from measured output duration, but only WAV/PCM were measurable — and mp3 is OpenAI's **default** speech format, so the default path always recorded $0 cost. Added an MPEG Layer III frame-header walk (no decoding, exact for CBR and VBR, skips ID3 tags). Verified: gateway-measured 5.208s matches ffprobe exactly.

### 3. `MONGODB_URL` database path is honored
`mongodb://host:27017/mydb` silently wrote to the hardcoded `gomodel` database — even though `.env.template` shows the path form. Resolution now lives in one place in the storage layer: explicit `MONGODB_DATABASE` > URL path > `gomodel`. Removed the duplicated defaulting and unused `storage.DefaultConfig`.

### 4. Unknown models return 404 `model_not_found`
Previously 400 `invalid_request_error`; OpenAI returns 404 with code `model_not_found`, and clients key on that. Four duplicated error constructions collapsed into `core.NewModelNotFoundError`. Fallback detection (substring `not_found`/`unsupported`) is unaffected.

### 5. Audio capability mismatch (Groq / Gemini)
Only the OpenAI provider implemented `AudioProvider`, yet `/v1/models` advertised `groq/whisper-large-v3` and Gemini TTS models — calling them could only fail with 400.
- **Groq** now serves audio: its provider was ~200 lines hand-duplicating `openai.CompatibleProvider`; it is now a thin delegating wrapper (same capability surface plus the two audio methods). Groq whisper transcription verified live.
- **Listing honesty:** `ListPublicModels` skips audio-only models (modes exclusively `audio_speech`/`audio_transcription`) when the owning provider has no audio support. Models without mode metadata stay listed — missing data must not hide a model.

## Provider-specific behavior
- Conversation history is converted through the existing input-union shapes, so it works with native Responses providers and chat-translated ones alike; stored item IDs are stripped before dispatch so upstreams don't treat them as item references.
- Groq keeps Responses-via-chat translation; its router capability surface is unchanged apart from audio.

## Test plan
- [x] New unit tests: Mongo DB-name resolution (8 cases), mp3 duration (10 cases incl. ID3/junk/MPEG-1), conversation turns (5 tests: merge/strip/append, 404, previous_response_id rejection, ref shapes, streaming append), audio capability filter
- [x] Updated tests encoding the old 400-on-unknown-model contract to the new 404
- [x] Full suite: 53 packages pass (`go test ./...`), pre-commit `make test-race` passed
- [x] Live verification against real providers: multi-turn + cross-provider + streaming conversations, Groq whisper STT, TTS cost row in usage DB matching ffprobe duration, Mongo URL-path database, 404 error shape

## Summary by CodeRabbit

* **New Features**
  * Conversation support for /v1/responses to merge history and append turns
  * MongoDB: MONGODB_DATABASE option to override DB from connection URL
  * Improved audio duration measurement with MP3 support for usage reporting

* **Bug Fixes**
  * Unsupported/unknown models now return 404 with code model_not_found
  * Audio-only models hidden for providers without audio support
  * MongoDB DB resolution now honors explicit setting and URL parsing

* **Documentation**
  * Clarified MongoDB configuration and DATABASE precedence in docs and template

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -192,6 +192,7 @@
 
 # MongoDB Configuration (recommended for high-volume logging)
 # MONGODB_URL=mongodb://localhost:27017/gomodel
+# MONGODB_DATABASE overrides the database named in MONGODB_URL (default: gomodel)
 # MONGODB_DATABASE=gomodel
 
 # =============================================================================
--- a/config/config.go
+++ b/config/config.go
@@ -82,9 +82,6 @@ func buildDefaultConfig() *Config {
 			PostgreSQL: PostgreSQLStorageConfig{
 				MaxConns: 10,
 			},
-			MongoDB: MongoDBStorageConfig{
-				Database: "gomodel",
-			},
 		},
 		Logging: LogConfig{
 			LogBodies:             true,
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -139,8 +139,8 @@ func TestBuildDefaultConfig(t *testing.T) {
 	if cfg.Storage.PostgreSQL.MaxConns != 10 {
 		t.Errorf("expected Storage.PostgreSQL.MaxConns=10, got %d", cfg.Storage.PostgreSQL.MaxConns)
 	}
-	if cfg.Storage.MongoDB.Database != "gomodel" {
-		t.Errorf("expected Storage.MongoDB.Database=gomodel, got %s", cfg.Storage.MongoDB.Database)
+	if cfg.Storage.MongoDB.Database != "" {
+		t.Errorf("expected Storage.MongoDB.Database to default empty (resolved by storage layer), got %s", cfg.Storage.MongoDB.Database)
 	}
 	if !cfg.Logging.LogBodies {
 		t.Error("expected Logging.LogBodies=true")
--- a/config/storage.go
+++ b/config/storage.go
@@ -33,9 +33,10 @@ type PostgreSQLStorageConfig struct {
 
 // MongoDBStorageConfig holds MongoDB-specific storage configuration
 type MongoDBStorageConfig struct {
-	// URL is the connection string (e.g., mongodb://localhost:27017)
+	// URL is the connection string; a database named in its path is honored
+	// (e.g., mongodb://localhost:27017/gomodel)
 	URL string `yaml:"url" env:"MONGODB_URL"`
-	// Database is the database name (default: gomodel)
+	// Database overrides the database named in the URL (default: gomodel)
 	Database string `yaml:"database" env:"MONGODB_DATABASE"`
 }
 
@@ -60,9 +61,6 @@ func (c StorageConfig) BackendConfig() storage.Config {
 	}
 	if cfg.SQLite.Path == "" {
 		cfg.SQLite.Path = storage.DefaultSQLitePath
-	}
-	if cfg.MongoDB.Database == "" {
-		cfg.MongoDB.Database = "gomodel"
 	}
 	return cfg
 }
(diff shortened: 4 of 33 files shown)


SantiagoDePolonia:
Pushed 26f0b24 addressing CI and review feedback:

**E2E failures** — the `tests/e2e` package is build-tagged (`//go:build e2e`) so it was missed by the local hook run; its three tests still encoded the old 400-on-unknown-model contract and now expect 404 `not_found_error`.

**Addressed review comments:**
- *Atomic conversation appends* (greptile): added `conversationstore.Store.AppendItems`, implemented atomically under the memory store's lock; `appendExchange` no longer does get-then-update. Covered by a race-detector test with 20 concurrent writers.
- *Nil-turn dereference* (greptile): `appendResponse` already had a nil-receiver guard so no panic was possible, but the guard is now explicit at the call site, which reads better.
- *Mongo DB name validation* (coderabbit): names parsed from the URL path now trim trailing slashes and reject multi-segment paths (db names cannot contain `/`); falls back to the default instead of failing at connect time. Tests added.
- *Truncated mp3 frames* (coderabbit) + *CodeQL allocation alert*: a final truncated frame is no longer counted as billable duration, and the flagged `len(history)+1` preallocation arithmetic is gone.
- *Handler store mutex* (coderabbit): conversation-store wiring now shares the handler mutex that already guarded `translatedSvc` (renamed `responseStoreMu` → `storesMu` to reflect that).
- *`error.code` assertions* (coderabbit): the 404 batch tests now assert `"code":"model_not_found"`.
- *Groq audio tests* (coderabbit): added `CreateSpeech`/`CreateTranscription` tests covering request translation (multipart fields, auth header), response passthrough, and an upstream-error path.

**Skipped:**
- *“Forward response lifecycle methods on Groq”* (greptile): intentional, not a regression — the previous hand-rolled Groq provider never implemented `GetResponse`/`CancelResponse`/etc. either, so the capability surface is unchanged. The wrapper deliberately delegates explicitly rather than embedding to avoid silently advertising native /responses lifecycle operations that this integration routes via chat translation. Can be revisited separately if Groq-native Responses support is wanted.

SantiagoDePolonia:
Pushed 8398b31 addressing the two follow-up comments:

- **Detach append context for non-streaming turns** (coderabbit): `appendResponse` now receives `context.WithoutCancel(ctx)`, mirroring the streaming observer, so a client disconnect right after provider success cannot cancel the store write under a future cancellation-aware store.
- **Tighten the concurrency test** (coderabbit): the test now decodes every stored item and asserts each writer id appears exactly once, so a duplicated-plus-lost pair can no longer slip past the length check.

PR #387, opened 2026-06-11
feat(aliases): add user_path restrictions to model aliases

vfeitoza:
Aliases can now be restricted to specific user paths. When a `user_paths` list is set on an alias, only requests whose `X-GoModel-User-Path` header matches one of the listed paths (or a descendant) will have the alias applied. Requests that do not match fall through to the original model name unchanged.

- An empty `user_paths` list (the default) applies the alias to all users
- A single "/" entry also applies the alias to all users
- Paths are matched exactly or by prefix (e.g. /team matches /team/alpha)
- Multiple paths can be specified per alias

Backend changes:
- `Alias.UserPaths []string` field added with `MatchesUserPath()` helper
- `Service.ResolveModelWithUserPath()` reads user_path from context and filters aliases that do not match the caller's path
- All three stores (SQLite, PostgreSQL, MongoDB) persist the new field; SQLite and PostgreSQL use a JSON column with a safe migration default
- `ModelResolver` interface updated to accept `context.Context` so the resolution chain can carry request-scoped user_path through to alias evaluation without changing public API signatures
- Admin handler accepts `user_paths` in the upsert request body
- `View.HasUserPathRestriction` flag exposed for dashboard display

Dashboard changes:
- Alias create/edit form includes a "User Paths" textarea (one path per line, reuses the same normalizer as model overrides)
- Existing alias data is populated in the form when editing
- Toggle and save operations preserve the current user_paths value

## Summary by CodeRabbit

* **New Features**
  * Admins can define per-alias user path restrictions via a new "User Paths (conditional)" textarea (one path per line; leave empty for all users).
  * Alias records now include an enabled flag and persist user-path rules so dashboard edits, toggles, and saves honor configured user-path restrictions.

### DIFF
--- a/internal/admin/dashboard/static/js/modules/aliases.js
+++ b/internal/admin/dashboard/static/js/modules/aliases.js
@@ -20,7 +20,8 @@
                 name: '',
                 target_model: '',
                 description: '',
-                enabled: true
+                enabled: true,
+                user_paths: ''
             },
             modelOverrideFormOpen: false,
             modelOverrideSubmitting: false,
@@ -142,7 +143,8 @@
                     name: '',
                     target_model: '',
                     description: '',
-                    enabled: true
+                    enabled: true,
+                    user_paths: ''
                 };
             },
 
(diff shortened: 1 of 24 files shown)


SantiagoDePolonia:
Thank you for your work! The PR inspired me to unify model overrides with aliases into a virtual model entity.

The underlying issue and the reason why this PR is needed, is that we treat aliases as something different from model overrides. Unifying those two entities into one is future-proof and helpful for maintenance. It also lets us simplify the UI a little bit.

Your work here was really helpful in detecting this underlying issue, but "good is the enemy of better".

Thank you again and the PR is coming!

PR #394, opened 2026-06-14
feat(realtime): add OpenAI Realtime voice via websocket reverse proxy

SantiagoDePolonia:
## What

Adds OpenAI-compatible **realtime (speech-to-speech / voice) support** to the gateway as a transparent websocket reverse proxy.

The core insight (after looking at how Bifrost, OpenAI, and Azure APIM do it): realtime is a **transport** concern, not a translation one. The provider's realtime event schema *is* the wire format, so the gateway's job is credential injection, routing, authorization, and observation — frames are relayed verbatim. No per-provider message translation.

## How

- **`core.RealtimeProvider`** — optional provider capability (like `AudioProvider`) returning a `RealtimeTarget{URL, Headers, Subprotocols}`. OpenAI implements it; the dial URL is derived from the configured base URL (`https://…/v1` → `wss://…/v1/realtime?model=`).
- **`internal/realtime`** — a small generic bidirectional pump (`github.com/coder/websocket`). Read limit raised to 16 MiB so base64 audio frames survive; dials upstream *before* upgrading the client so dial failures return a normal HTTP error.
- **Routes** — typed `GET /v1/realtime` (routes by `?model=`) and a websocket-upgrade branch on `/p/{provider}/v1/realtime`, both funneling through one resolution path (`RealtimeRouter.RealtimeTarget`).
- **Gating & observability** — auth, model-access, and budget checks reuse the same rules as the other model endpoints. Usage is tracked per `response.done` via a non-blocking inline tap (cheap `bytes.Contains` guard before parsing). Session lifecycle is logged.
- **Config** — `REALTIME_ENABLED` (default `true`). Only providers implementing realtime accept sessions (OpenAI today).

## Security

- Client credentials (master/managed key) authenticate at the gateway and are **never** forwarded upstream; provider credentials are injected at dial and **never** exposed to clients or logs.
- Websocket handshake headers are regenerated by the dialer (not forwarded).

## User-visible impact

Clients connect to `wss://<gateway>/v1/realtime?model=gpt-realtime-mini` with their gateway key and get a fully proxied OpenAI Realtime session — no provider API key on the client.

## Testing

- Unit tests: proxy relay / 512 KiB frame survival / dial-error-before-upgrade, URL + credential mapping (no key leak), usage extraction, router routing (model + provider hint), upgrade detection + upstream header merge.
- Verified **live** against OpenAI (`gpt-realtime-mini`) over both routes: full duplex round-trip (`session.created` → text deltas → `response.done`) and a persisted usage row; auth→401, unknown model→404, no model→400, non-realtime model→400. (The temporary live smoke test used for this was removed; only offline unit tests remain.)
- Gotcha worth noting: the GA realtime endpoint **rejects** the legacy `OpenAI-Beta: realtime=v1` header (`beta_api_shape_disabled`), so it is intentionally not sent.

## Docs

`REALTIME_ENABLED` documented in `.env.template`, `config/config.example.yaml`, and `CLAUDE.md`.

## Release Notes

* **New Features**
  * Added realtime (speech-to-speech) WebSocket endpoint at `/v1/realtime` with support for OpenAI, Bailian, and xAI providers.
  * Introduced `REALTIME_ENABLED` configuration option to control realtime endpoint availability.
  * Realtime sessions now support usage tracking with token counting.

* **Documentation**
  * Added configuration guidance for enabling/disabling realtime endpoints.

* **Tests**
  * Added comprehensive test coverage for realtime routing, WebSocket proxy, and provider implementations.


SantiagoDePolonia:
Thanks for the review — all five P1 findings were valid and are fixed in `3f811f3`:

1. **Passthrough allowlist bypass** (handlers.go) — the realtime websocket branch now checks `isEnabledPassthroughProvider` and returns the same `unsupportedPassthroughProviderError` as the HTTP passthrough path, so a provider disabled for `/p/{provider}/...` is unreachable via upgrade too.
2. **Workflow usage ignored** (realtime_service.go) — `usageTap` now also gates on `core.GetWorkflow(ctx).UsageEnabled()`, matching the other streaming paths.
3. **Beta header leaks upstream** (realtime_service.go) — `realtimeUpstreamHeaders` now strips a client-supplied `OpenAI-Beta` before dialing, so a stale `realtime=v1` can't turn a valid session into a dial failure.
4. **Audio costs miskeyed** (usage/realtime.go) — realtime audio/cached tokens now use the canonical `prompt_audio_tokens` / `completion_audio_tokens` / `prompt_cached_tokens` keys that `cost.go` prices, so they bill at audio rates (verified against `buildRawUsageFromDetails`).
5. **Default disables realtime on nil config** (http.go) — `handler.realtimeEnabled = cfg == nil || cfg.RealtimeEnabled`, consistent with the route-registration default.

Tests updated accordingly (header-strip and audio-key expectations); full suite + `make test-race` + `make lint` pass locally.

PR #406, opened 2026-06-16
fix(embeddings): honor encoding_format + fail loudly on empty ollama embeddings

SantiagoDePolonia:
## Summary

Two fixes that make `/v1/embeddings` robust against OpenAI-compatible local servers (LM Studio, Ollama, vLLM). Both were found while debugging a real report: `arabold/docs-mcp-server` → GoModel → LM Studio produced broken embeddings with no error.

## Fix 1 (primary): honor `encoding_format`

**Root cause, verified end-to-end with the actual LangChain client:**

- The OpenAI Python and JS/**LangChain** SDKs request `encoding_format: "base64"` **by default** and decode it client-side.
- **LM Studio ignores `encoding_format`** and always returns a float array.
- LangChain then mis-decodes the float array as packed bytes (`Buffer.from(arr, "base64")`), producing a corrupted vector with **1/4 the dimensions** (768 → 192). The downstream vector store breaks.
- This happens whether the client hits LM Studio **directly or through the gateway** — so it isn't introduced by GoModel, but the gateway is the only place that can repair it.

**Wire capture (real `@langchain/openai` `OpenAIEmbeddings`, the exact path `docs-mcp-server` uses):**

```
DIRECT to LM Studio : request encoding_format=base64 → response float array(768) → client decodes 192 ❌ CORRUPTED
VIA GoModel (after) : request encoding_format=base64 → response base64 string    → client decodes 768 ✅
```

The fix normalizes the embeddings response to the encoding the caller requested (Postel's Law): accept whatever the upstream returns, re-encode each vector as base64 (little-endian float32, matching OpenAI) or as a float array. Vectors already in the requested form, and payloads that parse as neither, are left untouched — so real OpenAI/base64 upstreams are a no-op.

- `internal/core/embeddings_encoding.go` — `NormalizeEmbeddingEncoding` + table-driven tests (round-trip fidelity, both directions, malformed-passthrough).
- `internal/providers/router.go` — apply it in `Router.Embeddings` using the client's requested `encoding_format`.

> Note: provider-prefix trimming in the response (`lmstudio/model` → `model`) was investigated as a suspect and **ruled out** — the LangChain client round-trips fine with a trimmed model name; it's shared with chat completions, which work.

## Fix 2 (secondary): fail loudly on empty ollama embeddings

When an OpenAI-compatible server is misconfigured as `type: ollama`, the adapter calls Ollama's native `/api/embed`, which LM Studio answers with HTTP 200 + an error body → zero embeddings → GoModel returned `200 {"data":[]}`, a silent failure with nothing in the logs. Now it returns a `502 provider_error` pointing the operator at the `openai`/`vllm` types.

- `internal/providers/ollama/ollama.go` + test.

## Docs

- `.env.template` and `config/config.example.yaml` — document configuring LM Studio as `openai`/`vllm` (it is OpenAI-compatible, **not** Ollama-compatible).

## User-visible impact

- OpenAI/LangChain embedding clients that default to base64 (e.g. `docs-mcp-server`) now get correct, full-dimension vectors through GoModel even when the upstream ignores `encoding_format`.
- Misconfigured ollama-typed embeddings fail with a clear, actionable error instead of silently returning empty results.
- No change for upstreams that already honor `encoding_format`, and no change to the chat path.

## Tests

- `internal/core`: `TestNormalizeEmbeddingEncoding`, `_RoundTrip`, `_Malformed`.
- `internal/providers/ollama`: `TestEmbeddings_NoVectorsErrors`.
- `make test-race` and `make lint` pass via pre-commit.

## Release Notes

* **New Features**
  * Added LM Studio provider support with OpenAI-compatible API integration
  * Added base64 encoding format support for embeddings (in addition to float arrays)

* **Bug Fixes**
  * Enhanced error detection for malformed embedding responses to provide clearer diagnostics

* **Documentation**
  * Added LM Studio configuration examples to setup guides

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -379,6 +379,17 @@
 # Set base URL to enable (default: http://localhost:8000/v1)
 # VLLM_BASE_URL=http://localhost:8000/v1
 
+# LM Studio (local, OpenAI-compatible server)
+# LM Studio speaks the OpenAI API (/v1/chat/completions, /v1/embeddings) and has
+# NO native Ollama API. Attach it to the openai type via a suffix so it gets a
+# descriptive name — here "openai-lmstudio". Do NOT use OLLAMA_BASE_URL for it:
+# that routes embeddings to Ollama's native /api/embed, which LM Studio doesn't
+# implement. The API key can be any non-empty value; LM Studio ignores it.
+# OPENAI_LMSTUDIO_BASE_URL=http://localhost:1234/v1
+# OPENAI_LMSTUDIO_API_KEY=lm-studio
+# Optional: pin the served models (LM Studio only exposes loaded ones).
+# OPENAI_LMSTUDIO_MODELS=text-embedding-nomic-embed-text-v1.5
+
 # Amazon Bedrock
 # Bedrock has no per-provider API key. Auth comes from the standard AWS
 # credential chain — env vars (AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY/
--- a/config/config.example.yaml
+++ b/config/config.example.yaml
@@ -267,6 +267,16 @@ providers:
   #   base_url: "https://api.example.com/v1"
   #   api_key: "..."
 
+  # Example: LM Studio. Despite being "Ollama-like", LM Studio speaks the
+  # OpenAI-compatible API (/v1/chat/completions, /v1/embeddings) and has NO
+  # native Ollama API. Configure it as "openai" (or "vllm"), NOT "ollama" —
+  # the ollama type sends embeddings to Ollama's native /api/embed, which LM
+  # Studio does not implement.
+  # lmstudio:
+  #   type: openai
+  #   base_url: "http://localhost:1234/v1"
+  #   api_key: "lm-studio"  # any non-empty value; LM Studio ignores it
+
   # Example: Groq (OpenAI-compatible)
   # groq:
   #   type: "openai"
--- a/internal/core/embeddings_encoding.go
+++ b/internal/core/embeddings_encoding.go
@@ -92,6 +92,11 @@ func base64ToFloatArray(raw json.RawMessage) (json.RawMessage, bool) {
 	if err := json.Unmarshal(raw, &encoded); err != nil {
 		return nil, false
 	}
+	// Reject oversized payloads before DecodeString allocates the decode buffer,
+	// not just after, so a pathological upstream can't force a large allocation.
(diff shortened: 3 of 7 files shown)


SantiagoDePolonia:
Addressed the automated review feedback in 6a235c94:

- **Greptile P1 (empty batch → misleading 502):** the ollama no-vectors guard now fires only when the request actually carried input (`embeddingInputIsEmpty`). An empty input batch passes through as an empty response instead of a "reconfigure LM Studio" error. Added `TestEmbeddings_EmptyInputNoError`.
- **CodeQL (allocation overflow at embeddings_encoding.go):** added `maxEmbeddingDims` (1<<20) guards before the `len(values)*4` sizing in both encode/decode helpers; oversized payloads are left untouched rather than allocated.
- **CodeRabbit (loose test assertion):** `TestEmbeddings_NoVectorsErrors` now asserts the full contract — nil response, `*core.GatewayError`, HTTP 502, and the actionable message — instead of just non-nil.

`make test-race` and `make lint` pass via pre-commit.

SantiagoDePolonia:
Addressed the second review round in f1711cd0:

- **CodeRabbit — `embeddingInputIsEmpty` misses typed empty slices:** added a `reflect` fallback so any empty slice/array (e.g. `[]string{}`), not just `[]any{}`, is recognized as an empty batch and won't trigger the 502. In practice `Input` is JSON-sourced (`string`/`[]any`/`nil`), but this hardens it for directly-constructed requests. Regression test now covers `[]any{}`, `[]string{}`, `""`, and `nil`.
- **CodeRabbit — pre-check base64 size before decode:** `base64ToFloatArray` now rejects payloads larger than `EncodedLen(maxEmbeddingDims*4)` before `DecodeString` allocates, complementing the existing post-decode dimension cap.

`make test-race` and `make lint` pass via pre-commit.

SantiagoDePolonia:
Addressed in 0a8e832f (Greptile P1 "Scalar inputs skip errors"):

Good catch — treating scalar `""` and `nil` as an "empty batch" reopened the silent-failure hole. `embeddingInputIsEmpty` now exempts **only** genuine empty arrays/slices (including typed slices like `[]string{}`); scalar and `nil` inputs stay on the loud-error path, so a misconfigured OpenAI-compatible endpoint returning a `200` error body for `""`/`null` still surfaces the `502` instead of being swallowed as `{"data":[]}`.

`TestEmbeddings_EmptyInputNoError` now asserts both halves: `[]any{}` / `[]string{}` pass through with empty data, while `""` and `nil` return the provider error. `make test-race` + `make lint` pass.

PR #408, opened 2026-06-17
feat(dashboard): support fully offline use

### LINKED ISSUE #407: Can dashboard supports offline use?
Currently, when using the dashboard, several js files, fonts and icons need to be loaded online. This results in the dashboard being completely unavailable in an offline environment.

SantiagoDePolonia:
Closes #407

## Problem
The admin dashboard loaded several assets over the network — the Inter typeface from Google Fonts and chart.js, Alpine.js, and Lucide icons from the jsDelivr CDN. In an offline environment the dashboard was effectively unusable: no styling, no charts, no interactivity, no icons.

### Vendor all front-end assets (offline support — the issue)
- **`static/vendor/`** — `chart.umd.min.js`, `alpine.min.js`, `lucide.min.js`, served from the embedded static tree via the existing `assetURL` pipeline. Each file was verified byte-for-byte against the SRI hashes previously declared in `layout.html`, and the trailing `//# sourceMappingURL=…` comments were stripped so browsers don't 404 on `.map` files that aren't shipped.
- **`static/fonts/`** — Inter vendored as one variable `woff2` per subset (`latin`, `latin-ext`) with a local `@font-face` stylesheet. Inter v20 ships a single variable file per subset, so one `@font-face` with `font-weight: 400 700` covers every weight the UI uses. System-font fallbacks already cover non-Latin text.
- `layout.html` now references these local assets and drops the `preconnect`/CDN/`integrity` attributes; the `//go:embed` directive is extended to include `static/vendor/*.js`, `static/fonts/*.css`, and `static/fonts/*.woff2`.
- No build system, bundler, or asset-serving changes — assets drop into the directory the embed/FileServer pipeline already handles.

### Fix a 400 surfaced by exercising the dashboard offline
With audit logging disabled (`LOGGING_ENABLED` is **off by default**), the Audit Logs page failed with `400 "invalid limit, expected positive integer"`. The disabled-reader fast paths for `/admin/audit/log` and `/admin/usage/log` returned `limit:0`; the dashboard reuses the returned `limit` as the next request's `limit` param, so the follow-up request sent `limit=0`, which validation rejects. Both fast paths now echo the requested `limit`/`offset`, matching the enabled-reader contract (which clamps `limit` to a positive value). The pages now load cleanly with their readers off, showing an empty list instead of erroring.

## Tests
- `dashboard_test.go`: `TestIndex_HasNoExternalResources` (page fetches no `http(s)` resources) and `TestStatic_ServesVendoredAssets` (vendored libs + fonts are embedded and served).
- `dashboard-layout.test.cjs`: updated to assert local font/vendor references and the absence of any CDN/preconnect.
- `handler_test.go`: `TestAuditLog_NilReader` / `TestUsageLog_NilReader` now assert the echoed `limit`.
- Full pre-commit suite green (`make test-race`, `make lint`, dashboard JS tests, gofmt/imports).

## Summary by CodeRabbit

* **New Features**
  * Admin log API responses now consistently include pagination metadata (limit, offset) even when no reader is available.

* **Bug Fixes**
  * Admin dashboard assets (fonts, vendor scripts, libraries) are now served locally via embedded/offline assets, removing external CDN dependencies.

* **Tests**
  * Added/updated tests to verify offline safety (no external resource references) and confirm required embedded assets are served correctly.

* **Documentation**
  * (None)


SantiagoDePolonia:
Thanks for the reviews. Addressed in bcccedc6:

- **Default page size when `limit` is omitted** (CodeRabbit ×2, Greptile ×2): both disabled-reader fast paths (`/admin/audit/log`, `/admin/usage/log`) now echo the documented default (25 / 50) via new `defaultAuditLogLimit` / `defaultUsageLogLimit` constants placed alongside the existing `maxAuditLogLimit` / `maxUsageLogLimit`, so the response matches the enabled-reader clamp for any caller, not just the dashboard.
- **Assert `offset` echo** (CodeRabbit): both nil-reader tests now omit `limit` (exercising the default path) and assert both the default `limit` and `offset`.
- **Protocol-relative URLs** (CodeRabbit): the offline-resource guard regex now also catches `//cdn…` forms (`(?:https?:)?//`).

Intentionally **not** changed:
- **`font-family: 'Inter'` quotes** (CodeRabbit / Stylelint): this repo has no Stylelint config and no CSS-lint step in pre-commit or CI, and the existing `static/css/dashboard.css` quotes `"Inter"` as well. Keeping the quotes matches the codebase convention; happy to drop them if a Stylelint rule is added later.

Full pre-commit suite (`make test-race`, `make lint`, dashboard JS tests, gofmt/imports) passes.

PR #414, opened 2026-06-20
feat(router): intelligent cost-aware + latency-aware routing

Barsit:
## Summary

When multiple providers serve the same model ID, the gateway now proactively selects the best provider based on a weighted cost + latency score — instead of the previous deterministic first-wins strategy where the first registered provider always handled the request.

## Architecture

```
Request: model="gpt-4o" (no provider pinned)
         |
         v
    resolveProvider()
         |
         +- providerHint != "" --> exact route (existing logic, unchanged)
         |
         +- providerHint == "" --> collectCandidates(modelID)
                                       |
                                       v
                                  StrategyRegistry.Resolve(ctx)
                                       |
                                       v
                                  RoutingStrategy.Select()
                                       |
                                       v
                                  Best candidate --> route
```

### New components (8 files)

Routing strategies under `internal/router/`:

| Component | File | Responsibility |
|-----------|------|---------------|
| RoutingStrategy interface | `strategy.go` | Pluggable strategy + ProviderCandidate |
| BalancedStrategy | `strategy_balanced.go` | Weighted cost+latency scoring (default) |
| CostOnlyStrategy | `strategy_cost.go` | Pure cost selection |
| LatencyOnlyStrategy | `strategy_latency.go` | Pure latency selection |
| FirstFitStrategy | `strategy_firstfit.go` | First acceptable (historical compat) |
| Scoring helpers | `scoring.go` | Normalization, filtering, tie-breaking |
| StrategyRegistry | `registry.go` | Name to factory map + override resolution |
| Context helpers | `context.go` | Context key helpers |

Supporting:

| Package | File | Responsibility |
|---------|------|---------------|
| `internal/llmclient/` | `latency.go` | EWMA latency/error tracker |
| `internal/core/` | `interfaces.go` (+ProviderStats) | Runtime stats interface |
| `internal/server/` | `routing_strategy.go` | X-GoModel-Routing-* header capture |
| `config/` | `router.go` | Router config struct + validation |

### Modified components

| File | Change |
|------|--------|
| `llmclient/client.go` | Integrate LatencyTracker, implement ProviderStats |
| `providers/registry.go` | Add ListProvidersForModel (O(k), registration-order) |
| `providers/router.go` | Insert strategy selection into resolveProvider |
| `providers/init.go` | Wire strategy registry at startup |
| `providers/strategy_init.go` | Build registry from RouterConfig |
| `gateway/inference_prepare.go` | Set RoutingEligible context flag |
| `config/config.go` | Add RouterConfig field |
| `config/config.example.yaml` | Document router section |
| `.env.template` | Document MODEL_ROUTING_* env vars |
| `CLAUDE.md` | Document intelligent routing summary |

## Key design decisions

1. **Default balanced strategy** (cost 0.6 + latency 0.4). Single-provider configs are unaffected (one candidate, no decision needed)
2. **Missing data scores 1.0** (worst), so complete-data providers are preferred
3. **Ties broken by ProviderName** lexicographic order for determinism
4. **Three single-purpose strategies**: cost_only, latency_only, first_fit
5. **Per-request header overrides**: `X-GoModel-Routing-Strategy` and `X-GoModel-Routing-Weights`; invalid values silently fall back (WARN log)
6. **Explicit provider bypasses routing**: `openai-east/gpt-4o` or `req.Provider` non-empty leads to exact route, no strategy involvement

## Bugs fixed during development

1. **Mutual recursion**: cost/latency single-factor strategies recursed infinitely when both dimensions were absent. Fixed by computing the fallback dimension's score inline
2. **Free model mis-classification**: normalizeCost treated cost=0 as unknown. Fixed by returning -1 sentinel from perMtokCost when pricing is absent
3. **Production path never executed strategy**: PrepareChatRequest set req.Provider via ResolveModel, then resolveProvider skipped strategy when providerHint was non-empty. Fixed via RoutingEligible context flag set during prepare for unqualified requests
4. **FirstFit order**: used alpha-sorted candidates instead of registration order. Fixed by implementing ListProvidersForModel returning entries in registration order

## Testing

- **38 unit tests** (router + llmclient): strategy selection logic, edge cases (empty candidates, open circuits, error rate boundary, equal scores, free model vs unknown, tiebreaks), registry resolution, EWMA convergence
- **8 integration tests** (providers): cheaper/faster selection, header override, invalid override fallback, explicit provider bypass, single-candidate no-op, weights override, FirstFit registration order
- **lint: 0 issues**, full project build passes
- All new tests + existing Router tests pass

## Summary by CodeRabbit

* **New Features**
  * Added **Intelligent Routing** to automatically select the best provider when multiple providers can serve the same model and no provider is pinned.
  * Supports `balanced` (cost/latency weighted) and `first_fit` (registration-order), with configurable `max_error_rate`.
  * Routing decisions can consider smoothed latency and error-rate signals, including circuit-breaker states.
  * Added per-request routing override via `X-GoModel-Routing-Strategy`.
* **Bug Fixes**
  * Requests that explicitly pin a provider (or use `provider/model`) bypass Intelligent Routing.
  * Invalid per-request routing override headers are ignored and the global strategy is used.
* **Documentation**
  * Added feature docs, config reference updates, and API example curl commands.


Barsit:
## Review findings assessment

### Fixed

| # | Finding | Source | Fix |
|---|---------|--------|-----|
| 1 | BalancedStrategy zero-weight defaulting bug (costW check used modified variable) | coderabbitai | Changed to check s.CostWeight instead of costW |
| 2 | CostOnlyStrategy degrades to latency when free model has pricing | coderabbitai, greptile | Added hasPricingData() helper, replaces max==0 sentinel |
| 3 | LatencyOnlyStrategy same issue | coderabbitai | Added hasLatencyData() helper |
| 4 | EWMA first-sample detection using value==0 treats genuine zero as uninitialized | coderabbitai | Added initialized bool field to ewma struct |
| 5 | ListProvidersForModel returns nil instead of empty slice | coderabbitai | Changed to return []ModelWithProvider{} |
| 6 | WithStrategyOverride stores empty strings | coderabbitai | Added empty-string guard |
| 7 | NaN/Inf accepted as valid config values | coderabbitai | Added math.IsNaN/IsInf checks in router config + ParseWeights |
| 8 | ParseWeights accepts trailing garbage (0.6x,0.4) | coderabbitai | Added full-string consumption check |
| 9 | Model-syntax provider overridden when req.Provider is empty | coderabbitai, greptile | Added slash-in-model check in resolveProvider eligibility |

### Assessed as not actionable

| Finding | Reason |
|---------|--------|
| hasProviderInModel misidentifies slash models | In GoModel, ANY slash is provider/ID syntax. Slash IDs go through proper fallback. Consistent with existing routing semantics. |
| Docstring coverage < 80% | Standard bot threshold; 45% is reasonable for established codebase. |
| Patch coverage 69.6% | Core strategy/scoring logic has 94% coverage. Lower coverage in wiring/init code that is hard to unit-test in isolation. |
| Audio/realtime routing | These paths do not set the eligibility flag, so strategy does not apply to them by design. Even if it did, callers do capability assertions. |
| Zero max_error_rate accepted | Zero is a documented sentinel for "use default" (0.5), consistent with the rest of the gateway's config patterns. |

Barsit:
Additional fix pushed:

- **ParseWeights trailing-zeros regression**: The `Sprintf` full-string consumption check I added earlier incorrectly rejected valid inputs like `"0.60,0.40"` (because `Sprintf("%v", 0.6)` produces `"0.6"`, not `"0.60"`). Fixed by using `Sscanf("%f%s")` with a trailing string variable — `n=1` means exactly the float was consumed (no trailing chars), while `n=2` means there's trailing garbage. Also expanded `TestParseWeights_Valid` to cover trailing zeros, whitespace, whole numbers, and zero-cost cases.

All previous findings (balanced zero-weight bug, free pricing degradation, NaN/Inf validation, EWMA init flag, model-syntax provider override, etc.) remain fixed in the squashed commit.

SantiagoDePolonia:
I think it's a good direction, but the PR is messy. There are files committed which shouldn't be there, and if you want it merged into GoModel it'll need some more work. Here are the findings worth improving:

1. **Drop the committed artifacts.** `gomodel.exe` + `gomodel.exe~` (Windows binaries, ~67MB each), `c.out`, `coverage-bailian*`, and the whole `.idea/` folder shouldn't be in the repo — please remove them and add to `.gitignore` (`*.exe`, coverage outputs, `.idea/`). Out of curiosity, how did the `.exe` end up in there?

2. **Don't reinvent provider parsing.** `hasProviderInModel` re-derives the provider prefix from the raw model string, but the request already goes through `ParseModelSelector` / `splitQualifiedModel`, which splits provider/model for you. It's also buggy — it treats any slash as a provider prefix, so legit slash model IDs like `meta-llama/Llama-3-70b` get wrongly marked ineligible for routing. Reuse the existing parsed selector instead.

3. **The abstraction is a bit much for what it does.** `balanced` is the only strategy most setups will use — `cost_only`/`latency_only` are just `balanced` with a zero weight. I'd keep `balanced` (+ `first_fit`) and drop the rest plus the header weight overrides until there's real demand. The interface itself is fine, just trimming surface.

4. **Latency tracking is always-on**, even for single-provider configs that can't benefit, and it's in-memory per-instance — so it resets on restart and isn't shared across replicas. Worth gating + a comment noting that.

5. **Check `gofmt`/lint** — the new `const` block in `config/router.go` and the `Router` field in `config/config.go` look misaligned, so I'd double-check the "0 lint issues" claim covers the new files.

Also heads up: some of the more advanced cost-optimization routing might end up in a pro / closed-source build down the line, so let's keep this core routing seam clean and pluggable rather than baking too much logic into it.

SantiagoDePolonia (changes requested):
The requested changes are described in my previous comment.

Barsit:
@SantiagoDePolonia 感谢审查，5 点均已修复：

### 1. 删除工件文件
已从 git 追踪中删除 `gomodel.exe`、`gomodel.exe~`、`c.out`、`coverage-bailian*`、`.idea/`，更新了 `.gitignore`。

### 2. 复用 provider 解析逻辑
删除了 `hasProviderInModel` 函数，直接使用 `core.ParseModelSelector`。该函数之前与 `splitQualifiedModel` 完全重复且注释有误导性。

### 3. 精简策略抽象
- 删除 `CostOnlyStrategy` 和 `LatencyOnlyStrategy`
- 只保留 `BalancedStrategy`（默认）和 `FirstFitStrategy`
- 删除 `X-GoModel-Routing-Weights` 头部及其对应的 `ParseWeights`、`WeightsOverride`、`WithWeightsOverride` 等全部代码
- `X-GoModel-Routing-Strategy` 头部保留（可在 `balanced`/`first_fit` 间按请求切换）

### 4. Latency tracking 条件控制
- 通过 `llmclient.SetLatencyTrackingEnabled(false)` 开关控制
- 单 provider 配置（`len(providerMap) < 2`）在启动时自动关闭
- `LatencyTracker` 的注释已写明：in-memory、重启清零、不跨副本共享

### 5. gofmt/lint
- `gofmt -w` 修复了所有对齐问题
- `golangci-lint` 0 issues
- pre-commit hook 已加入自动 `gofmt -w`，避免未来出现格式化不一致

Barsit:
@SantiagoDePolonia All 5 points have been addressed:

### 1. Remove committed artifacts
gomodel.exe, gomodel.exe~, c.out, coverage-bailian*, .idea/ -- removed from git tracking, added to .gitignore.

### 2. Don't reinvent provider parsing
Removed hasProviderInModel entirely. It was a duplicate of splitQualifiedModel with a misleading comment. Now uses core.ParseModelSelector directly.

### 3. Trim the abstraction
- Deleted CostOnlyStrategy and LatencyOnlyStrategy files
- Kept only BalancedStrategy (default) and FirstFitStrategy
- Removed X-GoModel-Routing-Weights header and all associated code: ParseWeights, WeightsOverride struct, WithWeightsOverride/WeightsOverrideFromContext, applyWeightsOverride
- Kept X-GoModel-Routing-Strategy header -- still useful for per-request switching between balanced and first_fit

### 4. Gate latency tracking
- Added llmclient.SetLatencyTrackingEnabled(bool) -- disabled at startup when len(providerMap) < 2
- Added nil-safe guards on all LatencyTracker accessors (Record, P50Latency, P99Latency, ErrorRate)
- Added WARNING comment on LatencyTracker documenting in-memory-only semantics (resets on restart, not shared across replicas)

### 5. gofmt/lint
- gofmt -w applied to all files -- alignment issues fixed
- golangci-lint: 0 issues
- Pre-commit hook updated to auto-run gofmt -w before lint

Barsit:
@SantiagoDePolonia The 5 points you raised have all been addressed:

1. ✅ Committed artifacts (gomodel.exe, c.out, .idea/) removed from git tracking, .gitignore updated
2. ✅ hasProviderInModel removed — now uses core.ParseModelSelector directly
3. ✅ Extra strategies (cost_only, latency_only) and X-GoModel-Routing-Weights header dropped
4. ✅ Latency tracking gated — disabled for single-provider configs via SetLatencyTrackingEnabled
5. ✅ gofmt/lint fixed globally, pre-commit hook runs gofmt -w automatically

Could you please take another look when you have a moment? Thanks!

.githooks/pre-commit
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
SantiagoDePolonia: This file definitely shouldn't be here.

Barsit:
@coderabbitai review feedback summary:

### Fixed (code pushed in this commit)

**Still-valid findings that were fixed:**

1. **scoring.go:176** — perMtokCost now returns -1 when Pricing exists but both InputPerMtok and OutputPerMtok are nil, so non-per-token pricing is correctly treated as unknown rather than free.
2. **latency.go:105** — Changed from `duration.Milliseconds()` to `float64(duration.Nanoseconds()) / 1e6` to preserve sub-millisecond precision.
3. **context.go:45** — WithStrategyOverride(ctx, "") now stores nil instead of returning ctx unchanged, so empty-string clears any prior override.
4. **routing_strategy.go:26** — Strategy name is lowercased before storing in context.
5. **strategy_init.go:27** — SetDefault error is now explicitly handled rather than silently discarded.
6. **http.go:291** — Comment updated to no longer reference the removed X-GoModel-Routing-Weights header.
7. **router_strategy_integration_test.go:75+** — NewRouter errors now checked via err return instead of discarded.
8. **router_strategy_integration_test.go:272-273** — zombie `http.StatusOK` import guard removed.
9. **strategy_test.go:40,52** — Error messages corrected: "expected azure" → "expected zeta".
10. **latency_test.go** — Added TestLatencyTracker_LeadingSuccessesDoNotSkewErrorRate regression test.
11. **config/router.go:63** — Trim-before-empty-check fixed.

### Already fixed in prior commits

12. **config/router.go:94** — NaN/Inf validation already present via math.IsNaN/IsInf checks.
13. **latency.go:22** — EWMA init already uses an `initialized bool` field.
14. **registry.go:725** — ListProvidersForModel already returns `[]ModelWithProvider{}`.
15. **strategy_balanced.go:45** — Zero-weight default already uses `s.CostWeight` (struct field) instead of `costW`.
16. **config/router.go:94** — NaN/Inf, trailing-garbage validation already in place.

### By design / not applicable

17. **router.go:317** — The `!strings.Contains(model, "/")` guard is correct per GoModel routing semantics: any slash in model is provider/model syntax, consistent with ParseModelSelector.
18. **client.go:248** — Latency tracker uses `err != nil` intentionally. CB failures that prevent dispatch also propagate err, so the error rate still captures them.
19. **strategy_cost.go/latency.go** — Both strategy files removed in the simplification.
20. **ParseWeights/WeightsOverride** — Feature removed in the simplification.
21. **inference_prepare.go (hasProviderInModel)** — Function removed, replaced by core.ParseModelSelector.
22. **strategy_test.go degradation tests** — Strategies removed, degradation path no longer exists.
23. **CLAUDE.md** — Documentation refined since.
24. **registry_test.go ParseWeights tests** — Feature removed.
25. **routing_strategy.go header rejection** — Weights header removed.
26. **strategy_test.go non-per-token pricing coverage** — Fixed by perMtokCost change.

All fixes tested and passing. Thanks for the thorough review!

README.md
```diff
@@ -145,11 +145,99 @@ GoModel is configured through environment variables and an optional `config.yaml
 
 ---
 
+## Intelligent Routing
```
SantiagoDePolonia: The feature description shouldn't be in readme

Barsit:
@SantiagoDePolonia All the feedback has been addressed and the PR has been cleaned up. Here's a summary of what was done:

### SantiagoDePolonia's 5 points
1. ✅ Committed artifacts (gomodel.exe, c.out, .idea/) removed from git tracking
2. ✅ hasProviderInModel removed — now uses core.ParseModelSelector
3. ✅ Extra strategies (cost_only, latency_only) and X-GoModel-Routing-Weights header dropped
4. ✅ Latency tracking gated — disabled for single-provider configs
5. ✅ gofmt/lint fixed globally, pre-commit hook runs gofmt -w automatically

### Additional inline comments addressed
6. ✅  — removed from PR (kept locally via .gitignore)
7. ✅  — feature description moved to , no feature docs in README

### coderabbitai findings addressed
8. ✅ perMtokCost returns -1 when both pricing pointers are nil (non-per-token pricing treated as unknown, not free)
9. ✅ Sub-millisecond latency precision preserved (float64 nanoseconds / 1e6)
10. ✅ WithStrategyOverride(ctx, "") properly clears prior override
11. ✅ EWMA initialization uses initialized bool (not value==0 sentinel)
12. ✅ NaN/Inf validation in config
13. ✅ Strategy name lowercased before context storage
14. ✅ Regression test for leading-success error-rate
15. ✅ Plus all other minor fixes and test improvements

### PR diff is now clean
- Only 27 files changed, all directly related to intelligent routing
- No unrelated upstream docs changes in the diff
- Based on latest upstream/main via rebase

Could you please take another look? Thanks!

SantiagoDePolonia (commented):
The architecture is sound and the interface/registry seam is genuinely the right abstraction (and the maintainer agrees). To get to merge:
1. Fix the auth bypass - revalidate the strategy-chosen selector against modelAuthorizer (or move selection before authorization). This is non-negotiable.
2. Gate routing eligibility so audio/realtime/other-capability endpoints can't be rerouted to an incapable provider.
3. Make latency-tracking enablement a per-router config, not a sticky package global.

Barsit:
@SantiagoDePolonia All three points have been addressed:

### 1. Auth bypass — fixed

Added a `ModelAccessValidator` callback to the Router (via `WithModelAccessValidator` option). After the strategy selects a provider, the callback re-validates the final selector against the model authorizer. If validation fails, the strategy's choice is rejected and the original provider is used with a WARN log.

### 2. Audio/realtime gating — analysed, no action needed

Traced the full execution path for audio, realtime, and embeddings — none of them go through `resolveProvider()`. They use their own provider resolution paths (`audio_service.go`, `realtime_service.go`, `PrepareEmbeddingRequest`) that don't set `RoutingEligible` and don't call into the router's `resolveProvider`. The eligibility gate in `router.go:315` already excludes these endpoints in practice.

### 3. Latency tracking config — fixed

Removed the package-level `latencyTrackingEnabled` global and `SetLatencyTrackingEnabled`. Added `LatencyTrackingEnabled bool` to `llmclient.Config`. `DefaultConfig()` now sets it to `true` by default. Single-provider setups or any code path that creates a client can now opt out at the config level rather than through a sticky global.

All three shipped as a squashed amend onto the same commit.

SantiagoDePolonia:
@Barsit
Sorry for not answering on time - I was overwhelmed by other tasks.

1. The idea here is great. Right now we need something similar for Failover - latency aware failover. If you want to work on it - feel free to create a PR.
2. This one is stale - a lot of things changed. I assume it's better to close it and optionally start work from the latest main.

Sorry again for belated response here.

BTW - feel free to share any feedback anytime

Barsit:
这是来自QQ邮箱的自动回复邮件。您的邮件我已收到。

PR #416, opened 2026-06-22
feat(virtualmodels): unify aliases and access overrides into virtual models

SantiagoDePolonia:
## Summary

Aliases (rename a new name → a real model) and access overrides (gate a scoped selector by `user_paths`) were stored, served, and surfaced as two separate subsystems despite being the **same operator concept**. The split duplicated `user_path` scoping and DB-migration handling and was the root cause of recent bugs.

This unifies them behind one entity — the **virtual model** — in a single `virtual_models` table keyed by `source`:

- a row **with `targets`** is a **redirect** (alias today; many targets + a strategy enable load balancing later),
- a row **without `targets`** is an **access policy** (scoped selector gated by `user_paths`).

Behavior is derived from `targets`-presence — there is **no `role` column** — and resolution stays **staged**: redirect runs early (resolver), the access gate runs late (authorizer), exactly as before.

See **`docs/adr/0008-virtual-models.md`**.

## Approach (low-risk composition)

To avoid re-porting subtle, tested resolution logic, the unified `virtualmodels.Service` **composes the existing `aliases` and `modeloverrides` engines** over the single store via role-partitioned adapters (`redirectStore` → `aliases.Store`, `policyStore` → `modeloverrides.Store`). The two engines run **verbatim**; only storage, wiring, and the service face are unified.

- **One-time idempotent migration** (`seedFromLegacy`) copies legacy `aliases` + `model_overrides` rows into `virtual_models` on first start when empty. Marked `REMOVE-LATER`.
- **Legacy tables left dormant** for one release (rollback safety); a later cleanup milestone drops them and the legacy packages.
- **`targets[]` + `strategy`** columns are present but inert → load balancing is a later additive change with **no re-migration**.
- **`app.go` net −14 lines**: two subsystems collapsed into one.

## Behavior preservation

v1 preserves today's behavior exactly:
- single-target redirect == today's alias; policy rows == today's overrides, enforced identically.
- `MODEL_OVERRIDES_ENABLED=false` still yields a nil authorizer (allow-all).
- existing `/admin/aliases` and `/admin/model-overrides` endpoints and the current dashboard keep working unchanged (fed the composed sub-services).

## Correctness guards

- **Cross-kind clobber rejected**: since `source` is one namespace, upserting a policy whose selector equals an existing alias name (or vice-versa) is rejected with a validation error rather than silently overwriting the other row.
- **Multi-target/strategy rejected** in `Service.Upsert` (v1) rather than silently dropped.

## Tests

New `internal/virtualmodels` suite: store round-trip (all fields), migration seed (copy + idempotency + source collision → redirect wins), unified resolution / access-gating / list+delete routing, and the two correctness guards. All existing `aliases`, `modeloverrides`, `app`, `gateway`, `server`, `admin` suites pass unchanged.

`go build ./...` ✅ · `go test ./...` ✅ · `gofmt` ✅ · `golangci-lint` 0 issues ✅

## Follow-up (not in this PR)

The unified `/admin/virtual-models` endpoint + single "Virtual Models" dashboard form is a frontend-focused change kept out of this PR to stay low-risk; the current UI already presents both on one page and is fully functional.

## Summary by CodeRabbit

* **Documentation**
  * Added an architectural decision record for the new unified “virtual model” concept and consolidated management approach.
* **New Features**
  * Introduced unified storage and admin listing/management for redirects (alias-style) and access policies.
* **Refactor**
  * Automatically migrates existing legacy redirect and access-override data on first startup.
* **Bug Fixes**
  * Prevents redirect and policy behaviors from interfering with each other during refresh and evaluation.
* **Tests**
  * Added unit and SQLite-backed coverage for migration idempotency, combined listing/deletion, and cross-type conflict handling.
* **Chores**
  * Removed the persisted model overrides enable/disable configuration option and updated docs/config examples/tests accordingly.


SantiagoDePolonia:
Thanks @greptile-apps — all four findings were valid and are addressed in `17d38e45`:

**P1 · Policy migration drops access (`seed.go`) — security.** Correct, and the most important one. On a source shared by a legacy alias and a legacy access override, the migration now **fails closed** with an actionable error (rename the alias or the override selector) instead of silently dropping the override. Silently losing an access control — which a provider-explicit request could then bypass — was the right thing to reject.

**P1 · Clobber guard races (`adapters.go`).** Correct — the two engines hold independent locks, so the `Get`→`Upsert` check-then-write wasn't atomic. The redirect/policy adapters now share a write mutex, serializing cross-kind upserts/deletes. Added a `-race` test run.

**P2 · Disabled overrides still refresh / poll (`runtime_refresh.go`, `factory.go`).** Correct parity regression. When `MODEL_OVERRIDES_ENABLED=false` the access-override engine is no longer built: `Overrides()` returns nil, so no background refresh goroutine starts, the gateway authorizer stays nil (allow-all), and `/admin/runtime/refresh` skips the `model_overrides` step — matching the old path that never initialized the subsystem.

New tests cover the fail-closed migration, the disabled-overrides path, and the cross-kind guard. `go test ./... -race` ✅ · `golangci-lint` 0 issues ✅

SantiagoDePolonia:
@coderabbitai good catch on the multi-replica angle. The shared mutex makes the check-then-write atomic **within a process** (the common single-instance case), but you're right that across replicas sharing one database it's best-effort.

I've chosen to document this as a known limitation rather than make it fully atomic, because:
- This is the **admin config plane** (operator actions, near-zero write concurrency), not the request hot path.
- The trigger is exotic: an alias named *identically* to an override selector, written concurrently to two replicas in the same window.
- The legacy two-table design couldn't hit this, but a fully atomic fix needs a **per-backend conditional upsert** (SQLite `json_array_length` / Postgres `jsonb_array_length` guards + 0-rows-affected detection + a Mongo filtered update) — meaningful complexity for a vanishingly rare admin-plane race.

If multi-replica admin-write safety becomes a real requirement, the conditional-upsert approach is the fix; noted in the code comment. Happy to take it on in a follow-up if maintainers prefer.

Re: docstring coverage — the package, exported types, and exported funcs are documented; the 32% figure is counting unexported helpers and test funcs.

PR #423, opened 2026-06-23
feat(virtualmodels)!: unify engine, admin API, and dashboard UI

SantiagoDePolonia:
## Summary

Finishes the virtual-models unification started in #416. Aliases and model access overrides are now one **virtual model** entity end to end: one native engine, one admin endpoint, one dashboard editor.

## Backend
- **One native engine, one in-memory index.** Redirect (alias) resolution and policy (access) gating run natively on `VirtualModel` rows behind a single atomically-swapped snapshot, one refresh loop. `internal/aliases` and `internal/modeloverrides` are deleted (logic ported; legacy tables kept one release as a rollback net, read only by a self-contained seed).
- **Authoritative `Enabled`.** A policy's `Enabled` governs access — disabled = off for everyone, enabled + `user_paths` = restricted, no row = `MODELS_ENABLED_BY_DEFAULT`. A single model can now be disabled.
- **Scoped redirects (covers PR #387).** `user_paths` on a redirect are enforced at resolution via the optional `gateway.UserPathModelResolver`; a scoped alias applies only to matching callers and otherwise falls through to the literal model name.

## Admin API (breaking)
- `GET/PUT/DELETE /admin/virtual-models` replaces `/admin/aliases` and `/admin/model-overrides`.

## Dashboard
- One virtual-model editor: Source locked + prefilled when editing an existing model, an always-present target field, `user_paths`, a 3-state Enabled/Restricted/Disabled switch, description.
- Per-row enable/disable toggle on every model; alias-like badge/row styling for any model that carries a virtual model; the pricing and virtual-model editors share one layout.

## Cleanups / perf
- Pre-parse redirect targets in the snapshot (no per-request `ParseModelSelector` on the hot path).
- Validate redirect `user_paths` instead of silently dropping them (a dropped path would widen a scoped alias to everyone).
- DRY the canonical-source lookup; tighten compile-time interface assertions.

## Verification
- `go build ./...`, `go vet ./...`, `go test ./...` — 57 packages green.
- `node --test …/*.test.cjs` — 318 JS tests green.
- Manual E2E against a running instance: alias create + `/v1/models` exposure, disable/enable a real model, and the PR #387 scoped redirect (matching `user_path` resolves to the target; others fall through to `model_not_found`).

ADR `docs/adr/0008-virtual-models.md` updated.

## Notes / follow-ups
- `docs/openapi.json` still lists `/admin/model-overrides`: `make docs-openapi` fails on a pre-existing swag parse error in `internal/server/messages_handler.go` (`json.RawMessage`), unrelated to this change, so the generated spec wasn't regenerated.
- `/v1/models` exposure for a scoped redirect is unscoped (listed for everyone); resolution/enforcement is correct. Filtering the listing by per-redirect `user_paths` is a small follow-up.
- Unifying the two batch-rewrite paths into one (with an access-validator hook) is a deferred DRY cleanup.

## Release Notes

* **New Features**
  * Unified virtual models management consolidates redirect and access policy administration into a single editor interface.
  * Admin API now uses `/admin/virtual-models` endpoint for managing model redirects and policies.

* **Documentation**
  * Architecture decision record updated with virtual models unification approach.


SantiagoDePolonia:
Thanks @greptile-apps — all three findings addressed in 8957e187:

- **Policy refresh slows down (P1)** — the unified refresh loop now uses the *shorter* of the model-cache and workflow intervals, restoring the policy side's faster cross-instance propagation (was lagging up to the model-cache interval).
- **Seed can stay partial (P1)** — collisions are now detected *before* any write, so a conflict can't leave `virtual_models` partially seeded (which the empty-table guard would then skip next startup). Test asserts the table is empty after a conflict.
- **Scoped aliases listed globally (P2)** — added `ExposedModelsForUserPath` (hides redirects scoped to `user_paths` the caller doesn't match); `/v1/models` prefers it via the optional `UserPathExposedModelLister`.

`go build`/`vet`/`golangci-lint`/`go test ./...` green (57 packages); added tests for the scoped-listing and the no-partial-seed behavior.

SantiagoDePolonia:
Addressed @coderabbitai's findings in ebcd3c08 (the two it overlapped with @greptile-apps — seed prevalidation and exposed-redirect user_path gating — were already in 8957e187):

**Fixed**
- **provider.go — request-time redirect rewrites now user-path aware** (Major/security): the shared Provider/BatchPreparer rewrite path used unscoped `ResolveModel`, so a `user_paths`-scoped alias could apply for callers outside its scope in **batch** and guardrail-executor requests (the main chat path was already scoped via the gateway). Threaded `ctx` and switched request-time rewrites to `ResolveModelForUserPath`; unscoped `ResolveModel` is kept only for inventory.
- **aliases.js:1072** — create-mode overwrite confirmation now runs even with an empty target (a policy upsert that could otherwise silently replace a row).
- **aliases.js:624** — `vmFormEffectiveEnabled` reflects the override's value, not its presence.
- **aliases.js:808** — `toggleRowEnabled` bails when virtual models are unavailable.
- **page-models.html:131** — Save is disabled while Remove is in progress.
- **seed_test.go** — asserts migrated `Enabled` semantics, incl. a disabled-alias case.

**Not changed (with reason)**
- **handler_usage.go:410** — pricing recalculation resolves an operator-supplied selector that has **no request user_path**, and v1 redirects have a single target, so unscoped resolution yields the correct concrete model for pricing. Using the user-path resolver there would wrongly fall through to the literal name.

`go build`/`vet`/`golangci-lint` (0 issues) / `go test ./...` (57 pkg) / `node --test` (318) all green.

SantiagoDePolonia:
Addressed the latest re-review findings in 81136a32:

- **`handlers.go` — exposure scoping decoupled from the authorizer** (@coderabbitai, Major): `UserPathExposedModelLister` now runs even when `modelAuthorizer` is nil (passing a nil target-allow), so a `user_paths`-scoped redirect is hidden from out-of-scope callers instead of falling back to unscoped listing.
- **`aliases.js` — enable toggle in default-disabled deployments** (@greptile-apps, P1): enabling a model only `DELETE`s a path-less policy when `default_enabled !== false`; otherwise it `PUT`s `enabled: true`, so the model is actually enabled rather than falling back to the disabled default while the UI says enabled.
- **`seed.go` — partial-write rollback** (@greptile-apps, P1): if a write fails mid-seed (DB error / context cancellation), already-written rows are rolled back (best-effort, fresh context) so a partial seed can't trip the `len(existing) > 0` guard and skip importing the rest next startup. Test added (`TestSeedFromLegacy_RollsBackPartialWriteOnError`).

`go build`/`vet`/`golangci-lint` (0 issues) / `go test ./...` (57 pkg) / `node --test` (318) all green.

SantiagoDePolonia:
Addressed the latest re-review in 32fb69e5:

- **`provider.go` — scope listing leak** (P1/security): `Provider.ListModels` merged exposed redirects with the unscoped `ExposedModels()`, leaking `user_paths`-scoped redirect IDs to callers outside their scope. It has `ctx`, so it now uses `ExposedModelsForUserPath(core.UserPathFromContext(ctx), nil)` — parity with the server `/v1/models` path. (Applied your suggestion.)
- **`seed.go` — atomic write** (P1): the legacy seed now writes **atomically** on the SQL backends. `SQLiteStore`/`PostgreSQLStore` implement an optional transactional `UpsertAll`, and the seed prefers it, so a failed seed leaves the table untouched rather than partially populated (no `len(existing) > 0` suppression). MongoDB — no cross-backend transaction without a replica set — keeps the per-row best-effort rollback fallback. Added `TestSQLiteStore_UpsertAllIsAtomic` (success commits all; a cancelled-context batch writes nothing).

`go build`/`vet`/`golangci-lint` (0 issues) / `go test ./...` (57 pkg) / `node --test` (318) all green.

On the residual MongoDB case: a fully atomic Mongo seed needs a session transaction (replica set), which isn't always available; the best-effort rollback + the fact that a seed error fails startup (so the process won't serve partial data on that run) is the pragmatic floor for this REMOVE-LATER migration. Happy to add a Mongo session-transaction path if you'd prefer.

SantiagoDePolonia:
Confirmed the type question from the review on the `factory.go` cadence fix: `config.Workflows.RefreshInterval` is a `time.Duration` (config/workflows.go:9, default `time.Minute`), while `Cache.Model.RefreshInterval` is `int` seconds (config/cache.go:20). The fix converts the seconds field (`* time.Second`) and uses the workflow field as-is, so `min(modelInterval, policyInterval)` compares like units — correct (and it mirrors the pre-collapse code). No change needed.

SantiagoDePolonia:
Addressed @coderabbitai's review of the UpsertAll work (19e6c093):

- **store_postgresql.go — fresh rollback context**: `PostgreSQLStore.UpsertAll` now defers `tx.Rollback` with a fresh `rollbackContext()` instead of the (possibly-canceled) seed context, matching the pattern in service.go. (SQLite's `tx.Rollback()` takes no context, so it's unaffected.)
- **store_test.go — real mid-batch rollback**: replaced the pre-`BeginTx` cancelled-context case (which never exercised rollback after a successful row) with a true mid-batch failure — row 1 is written inside the transaction, row 2 fails to encode (non-finite `Weight`), and the test asserts the table is left empty.

Also in this push (separate commit 3b16799f, the requested cleanups): unified the refresh cadence on `Cache.Model.RefreshInterval`, renamed the dashboard controller `aliases.js`→`virtual-models.js` (+ factory/flag collapse), and removed the dead `model-access-state-badge` CSS.

`go build`/`vet`/`golangci-lint` (0 issues) / `go test ./...` (57 pkg) / `node --test` (318) all green.

PR #428, opened 2026-06-25
feat(dashboard): add token cache meter and overview refinements

SantiagoDePolonia:
## Summary

Adds a **cache meter** to the dashboard overview plus several overview refinements. The meter answers *"what share of what we send is actually cached?"* — something the existing absolute Cache Hits / Local Cache cards didn't show.

## What's included

- **Tokens cache meter** — a 100%-stacked, input-token bar split into **regular (uncached) / locally cached / prompt cached**, with on-bar percentages, a legend key, and an empty state. Colors are on-palette (GoModel brown / emerald `--success` / lime `--prompt-cache-color`) and theme-correct.
- **Provider prompt-cache split in `UsageSummary`** — new additive fields `uncached_input_tokens` / `cached_input_tokens` / `cache_write_input_tokens`, aggregated per row via a **shared Go fold that reuses `EntryInputSegments`** across the SQLite/PostgreSQL/MongoDB readers (the provider quirks — field-name coalescing, Anthropic additive-vs-subset accounting — stay in one place rather than being re-encoded in SQL).
- **Total Requests now includes local cache hits** (with a `provider + cache` breakdown tooltip), computed on the frontend so the uncached-mode summary the meter relies on stays clean.
- **Per-provider docs link** — provider cards show the standard `?` help icon linking to `…/docs/providers/<slug>`, but only for provider types that have a docs page (allowlist derived from `docs/providers/*.mdx`; handles `opencode_go` → `opencode-go`).
- **Moved the Daily/Weekly/Monthly/Yearly switch onto the usage chart header** (it only affects that chart's grouping).
- **Reordered overview cards** so **Cache Hits** sits before **Estimated Cost**.
- Regenerated the API reference for the new `UsageSummary` fields.

## Provider-specific behavior

The prompt-cache split is normalized across providers via the existing `EntryInputSegments`: Anthropic reports cache reads/writes as additive to `input_tokens`, while OpenAI/Gemini/etc. report them as a subset — the aggregate reconciles both so the meter is accurate per provider.

## Testing

- `go build ./...`, `go vet`, gofmt clean; all Go tests pass — including a SQLite cross-check that the aggregated split equals `Σ EntryInputSegments` over a mixed-provider fixture (and catches the invalid sum-per-field shortcut), plus handler tests for the new JSON fields.
- All dashboard JS module tests pass (added coverage for the meter math, empty state, percentages summing to 100, Total-Requests-with-hits, and the provider docs link).

## Summary by CodeRabbit

* **New Features**
  * Added cache-aware usage breakdown fields (uncached, cached, and cache-write input tokens) to the usage API schema.
  * Added a Cache Meter to the Usage Overview (token-segment visualization, legend with percentages/tokens, and empty state).
  * Added provider documentation help (“?” links) where provider-specific docs are available.
* **Bug Fixes**
  * Prevented stale cache hits and cache-meter data after failed refreshes or period/range changes.
* **Documentation**
  * Updated Swagger/OpenAPI to expose the new cache-split fields.
* **Tests**
  * Expanded API/UI/reader tests for cache splitting, meter calculations/rounding, provider doc URL logic, and failure-state handling.

### DIFF
--- a/cmd/gomodel/docs/docs.go
+++ b/cmd/gomodel/docs/docs.go
@@ -6915,6 +6915,12 @@ const docTemplate = `{
         "usage.UsageSummary": {
             "type": "object",
             "properties": {
+                "cache_write_input_tokens": {
+                    "type": "integer"
+                },
+                "cached_input_tokens": {
+                    "type": "integer"
+                },
                 "total_cost": {
                     "type": "number"
                 },
@@ -6934,6 +6940,9 @@ const docTemplate = `{
                     "type": "integer"
                 },
                 "total_tokens": {
+                    "type": "integer"
+                },
+                "uncached_input_tokens": {
                     "type": "integer"
                 }
             }
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -9051,6 +9051,12 @@
       "usage.UsageSummary": {
         "type": "object",
         "properties": {
+          "cache_write_input_tokens": {
+            "type": "integer"
+          },
+          "cached_input_tokens": {
+            "type": "integer"
+          },
           "total_cost": {
             "type": "number"
           },
@@ -9070,6 +9076,9 @@
             "type": "integer"
           },
           "total_tokens": {
+            "type": "integer"
+          },
+          "uncached_input_tokens": {
             "type": "integer"
           }
         }
--- a/internal/admin/dashboard/static/css/dashboard.css
+++ b/internal/admin/dashboard/static/css/dashboard.css
@@ -18,6 +18,15 @@
   --danger: #ef4444;
   --prompt-cache-color: #40a811;
   --prompt-cache-color-bg: color-mix(in srgb, var(--prompt-cache-color) 14%, var(--bg-surface));
+  /* Cache meter segments. Declared once as semantic aliases: --accent/--info
+     are re-themed per mode, and var() resolves lazily per element, so these
+     stay theme-correct without duplicating across the light/dark blocks. */
+  --cache-meter-uncached: var(--accent);
+  /* Two distinct greens: a deep emerald (darkened --success) for local hits,
+     lime --prompt-cache-color for provider prompt-cache reads. --success is the
+     same in both themes, so the mix stays theme-consistent. */
+  --cache-meter-local: color-mix(in srgb, var(--success) 58%, #000);
+  --cache-meter-prompt: var(--prompt-cache-color);
   --sidebar-width: 240px;
   --radius: 8px;
   --chart-grid: #2a2826;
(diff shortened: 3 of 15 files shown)


SantiagoDePolonia:
Thanks @coderabbitai and @greptile-apps. Dispositions:

**Fixed** (5ce08f8a):
- a11y: `role="group"` + `aria-pressed` on the interval picker.
- Direct unit tests for `foldInputSegments` (happy path, malformed `raw_data`, scan error, iteration error).
- nil-reader summary test now asserts the cache-split keys are present.

**Won't fix (intentional):**
- **Two-pass snapshot** (sqlite/pg/mongo): real but cosmetic for a 30-day dashboard aggregate — nothing displayed ties the split to the totals (the meter uses the split fields; `uncached+cached+write ≠ total_input` for Anthropic by design), and any sub-ms skew self-heals on the next load. A read-transaction across 3 backends or a full-scan single pass isn't worth it here.
- **Mongo malformed `raw_data`**: pre-existing behavior (`GetUsageLog` decodes `raw_data` the same way); it's always written as a JSON object, and decoding into `any` would yield `bson.D` and silently drop the split for *all* Mongo rows.
- **PG/Mongo parity tests**: no live-DB harness in this package. The shared split logic is identical across backends and is now covered by the `foldInputSegments` unit tests above; only the per-driver scan differs.

SantiagoDePolonia:
Good catch @greptile-apps — valid P1. Fixed in ab86ba29: `fetchUsage` now resets `cacheOverview` alongside `summary` on a failed refresh, so Total Requests and the cache meter can't show the previous period's cache hits next to an empty summary. Added a regression test.

SantiagoDePolonia:
Valid, @coderabbitai — fixed in dc0bf5a7. The success path now clears `cacheOverview` before the first render (when the overview is visible), so a range switch no longer briefly mixes the new summary with the previous period's cache data; `fetchCacheOverview()` then repopulates it. Added a regression test.

SantiagoDePolonia:
Thanks @coderabbitai:

- **Exception path leaves stale data** — valid. Fixed in 82bc86b9: the `fetchUsage` catch now resets `summary`/`daily`/`cacheOverview` and re-renders on a `fetch()`/`json()` rejection, matching the handled HTTP-failure branch. Added a regression test.
- **Use `workflowCacheVisible()` for cache-hit logic** — skipping. `workflowCacheVisible()` governs whether the *workflow editor* offers a cache toggle (defaults visible when Redis/semantic infra is present). It's the wrong predicate for cache *analytics*: `/admin/cache/overview` hard-gates on `CacheEnabled == "on"`, so without `CACHE_ENABLED` there's no cache data to show. `cacheAnalyticsEnabled()` already mirrors that gate and matches the existing Cache Hits / Local Cache cards (`x-show="cacheAnalyticsEnabled()"`); broadening it wouldn't surface data and would make the meter inconsistent with those cards.

PR #429, opened 2026-06-26
docs(benchmark): add AWS gateway benchmark and refresh benchmarks page

SantiagoDePolonia:
## What

A reproducible AWS benchmark comparing **GoModel** against **LiteLLM, Portkey, and Bifrost**, plus the refreshed docs benchmarks page. This PR is the **benchmark core only** - the write-up and co-located tooling are split into draft **#430**.

`docs/2026-06-25_aws_gateway_benchmark/`:
- **`RESULTS.md`** - the documented results.
- **Harness** - one command (`run.sh`): builds the GoModel image, provisions a small AWS box (Terraform), runs all four gateways against a shared instant **mock** backend (so numbers reflect gateway overhead, not model latency), collects results, and tears the infra back down. Tooling under `remote/` (Go loadgen + mock, gateway configs, compose), `terraform/`, `scripts/summarize.py`.
- Updated `docs/about/benchmarks.mdx` to this June 2026 run (with a paid-AWS cost note).

## Results (c7i.large, shared mock backend)

| Gateway | p99 (ms) | Throughput | Peak RAM | Image (compressed) | Cold start |
|---|--:|--:|--:|--:|--:|
| **GoModel** | **6.9** | **4,900 req/s** | **37 MB** | **16 MB** | **0.56 s** |
| Bifrost | 18.3 | 3,100 | 143 MB | 77 MB | 7.1 s |
| Portkey | 30.5 | 950 | 112 MB | 59 MB | 1.1 s |
| LiteLLM | 39.3 | 324 | 2.3 GB | 372 MB | 25.5 s |

## Split out → draft #430
The benchmark write-up (`ARTICLE.md`, `cover.png` + `make_cover.py`, `charts/`) and the co-located `qa/` and `translation/` tooling are in **#430** to keep this PR focused.

## Not committed
Terraform state, the provider cache, the generated SSH key, and raw run output are gitignored.

### DIFF
--- a/docs/2026-06-25_aws_gateway_benchmark/.gitignore
+++ b/docs/2026-06-25_aws_gateway_benchmark/.gitignore
@@ -1,0 +1,9 @@
+# Benchmark outputs and local Terraform state / secrets — never commit.
+output/
+remote/results/
+terraform/.terraform/
+terraform/.terraform.lock.hcl
+terraform/*.tfstate
+terraform/*.tfstate.*
+terraform/bench_key.pem
+*.tar.gz
--- a/docs/2026-06-25_aws_gateway_benchmark/README.md
+++ b/docs/2026-06-25_aws_gateway_benchmark/README.md
@@ -1,0 +1,122 @@
+# AWS gateway latency & resource benchmark — GoModel vs LiteLLM vs Portkey vs Bifrost
+
+A reproducible, one-command benchmark that provisions a free-tier AWS instance,
+runs four AI gateways through identical workloads against a deterministic mock
+backend, measures latency and resource cost, and tears the infrastructure down.
+
+Because every gateway talks to the **same local mock backend**, the numbers
+reflect *gateway overhead*, not upstream model latency or network jitter.
+
+## What it compares
+
+Four OpenAI-compatible gateways, each pointed at the mock:
+
+| Gateway  | Image                                | How it reaches the mock |
+|----------|--------------------------------------|-------------------------|
+| GoModel  | built from this repo (`Dockerfile`)  | `OPENAI_BASE_URL` env   |
+| LiteLLM  | `ghcr.io/berriai/litellm:main-stable`| `configs/litellm-config.yaml` |
+| Portkey  | `portkeyai/gateway:latest`           | `x-portkey-custom-host` header (+ `TRUSTED_CUSTOM_HOSTS=mock`) |
+| Bifrost  | `maximhq/bifrost:latest`             | `configs/bifrost-config.json` (`network_config.base_url` + `allow_private_network`) |
+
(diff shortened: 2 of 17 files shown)


SantiagoDePolonia:
**Review triage — fixed the harness/test issues that don't affect the measurements; flagging the ones that do.**

Addressed in `857857f2` here (and the companion QA + translation fixes in #430): the security, robustness, lint, docs, and reproducibility items — explicit `ssh_ingress_cidr` + CIDR validation (no `0.0.0.0/0` default), IMDSv2 + encrypted root volume, Docker Compose plugin `sha256` verification, base-image digest pins, tracked Terraform lockfile, quoted `rsync -e`, loadgen flag validation, expanded mock redaction + a recording-mode data-race fix, `run-on-instance.sh` guards, `summarize.py` lint, the inaccurate "free-tier" wording, and the benchmarks page now links the published write-up.

The points below **touch the measurement itself** — fixing any of them would change the published numbers and/or require re-running the benchmark, so I left them as-is for an explicit decision rather than silently altering results:

1. **Loadgen leaves the SSE body undrained on early return** — `remote/bench-tools/loadgen/main.go` (~303). Returning at the terminal marker / idle gap without draining + closing the body blocks HTTP keep-alive reuse, so streaming requests can pay per-request connection setup. → affects **streaming latency/throughput**.

2. **Resource sampler loop is unthrottled** — `remote/run-on-instance.sh` (~176). The `docker stats --no-stream` loop has no delay; sampler CPU competes with the gateway under load. → affects **CPU%** and the **req/s-per-CPU efficiency** figures.

3. **`gw_model` isn't dialect-aware for Bifrost's Anthropic route** — `remote/run-on-instance.sh` (~89). Bifrost always gets `openai/gpt-4o-mini`, even on the `/anthropic/v1/messages` workload. → could change **Bifrost's measured behavior/coverage** on the messages dialect.

4. **`summarize.py` discovers targets from `runs[0]` only** — `scripts/summarize.py` (~137). A target/workload present only in a later trial would be dropped or aggregated inconsistently. → affects **which rows appear** and the **median/spread aggregation**.

5. **Peer gateway images aren't pinned** — `remote/docker-compose.yml` (~88). LiteLLM/Portkey/Bifrost are pulled by floating tags, so the exact versions measured aren't locked. (The translation harness in #430 was pinned to the digests from the last run; this benchmark compose still floats.) → **reproducibility of which software was compared**.

6. **Terraform resolves the latest AMI at apply time** — `terraform/variables.tf` (~28). The OS image can drift run-to-run. → **environment determinism / run-to-run comparability**.

Happy to change any of these — but each implies the headline numbers should be re-measured on a fresh run, so I kept them out of the "no-impact" cleanup.

PR #479, opened 2026-07-04
feat: public extension API and embeddable bootstrap

SantiagoDePolonia:
## What this does

Gives GoModel a small, feature-neutral extension surface so custom gateway binaries can be built on top of it — the open-source binary itself becomes a thin wrapper over the same API. Split out of #457 so the module-path rename can be decided independently; **this PR does not touch the module name** (434 rename-only files dropped from the original diff).

### 1. Public `ext` package + ingress wiring
A registry where custom builds register:
- **Request rewriters** — receive the raw JSON body of `POST /v1/chat/completions`, `/v1/messages`, `/v1/responses` (subroutes excluded) after auth and before workflow resolution, so body rewrites (including `model`) affect routing, failover, guardrails, budgets, and caching. Fail-closed, errors rendered in each endpoint's native dialect.
- **Echo middleware** (before gateway auth), **extra routes**, and **auth-skip paths**.

An empty registry adds zero request overhead — the middleware isn't even registered, so the default build is byte-for-byte unchanged in behavior.

### 2. Public `run` package
`run.Run(ctx, run.Options)` is the full importable gateway lifecycle (CLI, probes, dotenv, logging, config, providers, signals); `Options.Setup` fires only when the process actually serves, keeping `--version` and Docker healthchecks silent. `cmd/gomodel` keeps byte-identical CLI behavior.

### 3. Audit request-revision chain
`request_body` stays the original client request; a new `request_revisions` array records each ingress rewrite (rewriter, sizes, optional detail, rewritten body when body logging is on) — mirroring the provider-attempts pattern — and the dashboard renders them as "Rewritten" tabs.

The extension/embedding guide (`docs/extensions.md`) ships with #457, since the embedding recipe only becomes practical once the module is go-gettable. Until then the `ext`/`run` packages should be treated as experimental with no compatibility promise.

## Testing
Full unit suite, e2e, 402 dashboard JS tests, and golangci-lint all green. New table-driven tests cover registry semantics, per-endpoint gating, rewriter chaining, fail-closed error envelopes per dialect, >64KB/>1MB bodies, audit-keeps-original for streaming and buffered paths, revision recording with/without body logging, and run CLI behavior.

## Summary by CodeRabbit

* **New Features**
  * Added gateway extension APIs for registering request rewriters, extra middleware/routes, and auth-skip public paths.
  * Added request-rewrite audit revision tracking and new “Rewritten” panes in the audit dashboard.
  * Refined startup by consolidating lifecycle handling into the reusable runtime entry point.
* **Bug Fixes**
  * Request-rewrite failures now fail closed with client-facing rejection responses and preserve the original request in audit logs.
* **Tests**
  * Expanded unit and middleware/audit/UI tests, plus end-to-end CLI/runtime coverage.


SantiagoDePolonia:
Addressed the CodeRabbit review — disposition of each finding:

**Fixed:**
- **Credential headers in `ext.Input.Header`** — now redacted (`Authorization`, cookies, API keys via `core.IsCredentialHeader`) before rewriters see them; rewriters run post-auth and get `UserPath` for identity, so raw credentials are never needed. The `Result.Detail` contract now states it must never contain secrets. Core cannot meaningfully filter arbitrary rewriter-provided JSON beyond that — rewriters are compiled-in code at the same trust level as core — so redacting the input is the effective guarantee: secrets can't be echoed into audit details if the rewriter never receives them. Covered by a new assertion in the rewrite test.
- **`app.New` Extensions wiring untested** — extracted the four-field copy into `applyExtensions(serverCfg, registry)` and added a table test covering population and the nil-registry no-op (a full `app.New` bootstrap test would require provider/storage scaffolding disproportionate to four field copies).
- **`defaultProviderFactory` untested** — added a test asserting all 19 provider types are registered (via `RegisteredTypes()`) with metrics enabled and disabled. The Prometheus-hooks branch itself isn't directly assertable (`hooks` is unexported in `providers`); the test pins the registration list, which is the regression that matters when providers are added.
- **`--health`/`--ready` dispatch untested** — added `TestRunHealthAndReadyDispatch`: both flags probed end-to-end against a live stub gateway (nil error) and an unreachable port (error surfaced, exit code 1).
- **MD028 blockquote in `docs/extensions.md`** — fixed, but note the file moved to #457 (the embedding guide lands with the canonical module path); the two quote blocks are now a single blockquote with a `>` separator.

**Skipped:**
- **Early exit in `RequestRewriteMiddleware` when rewriters are empty** — not reachable: the middleware is only registered when `len(cfg.RequestRewriters) > 0` (`internal/server/http.go`), so an empty-slice guard inside it would be dead code; the zero-rewriter build pays nothing today.

Full suite re-verified: unit, e2e, dashboard tests, golangci-lint 0 issues.

SantiagoDePolonia:
@coderabbitai continue

PR #480, opened 2026-07-04
fix(resilience): circuit breaker correctness, stale provider inventory, fast health rechecks

SantiagoDePolonia:
## Summary

Review of the circuit breaker feature and the provider-health machinery around it, prompted by a user report of a down provider showing "Last checked 20m ago" with no way to control the cadence. Two confirmed breaker bugs fixed, the registry's failure handling reworked, and the previously invisible/undocumented pieces surfaced.

### Circuit breaker fixes (`internal/llmclient`)

- **Permanently wedged breaker**: a local request-build error (or caller cancellation) during the half-open state consumed the single probe slot without ever returning it — every subsequent request got `503 circuit breaker is open` until process restart. The probe slot is now released when a request ends without a provider verdict.
- **Client aborts counted as provider failures**: canceled contexts during stream establishment charged `RecordFailure`, so a handful of impatient clients could open the circuit for a healthy provider. Caller cancellations are now neutral; client deadlines still count.
- Dead code removed (`Allow()`, duplicate `getBaseURL()`), `NewWithHTTPClient` delegates to `New`, `extractStatusCode` uses `errors.As`.

### Provider health rework (`internal/providers`, `internal/virtualmodels`)

- **Stale inventory instead of the wipe**: a failed refresh keeps the provider's previous models, marked stale. Direct requests still resolve and fail at the provider with an honest 502/503 (manual failover rules can fire) instead of a misleading "model not found". New `Catalog.ModelAvailable()` keeps virtual-model load balancing skipping stale targets, and bare model IDs served by several providers prefer a healthy one — the same routing the wipe produced, without the data loss. The dashboard's previously unreachable "Degraded — previous inventory is still available" state is now what a down provider shows.
- **Parallel refresh sweep**: provider fetches ran sequentially under one shared 30s budget, so one slow upstream starved every provider after it (starved providers were recorded failed and lost their models).
- **Fast recheck of failed providers**: `PROVIDER_RECHECK_INTERVAL` (default 60s, `0` disables) re-probes only the providers whose last refresh failed, so outage recovery and the dashboard "Last checked" no longer wait for the next full refresh (`CACHE_REFRESH_INTERVAL`, default hourly).

### Observability & dashboard

- New `gomodel_circuit_breaker_state{provider}` gauge (0=closed, 1=half-open, 2=open) — previously breaker state was invisible everywhere.
- Dashboard "Last checked" shows the most recent of model-fetch and availability-check timestamps.

### Docs

`resilience.mdx` now documents what counts as a retryable vs breaker failure (429/4xx/cancellation exemptions), per-path retry behavior (streaming never retries; passthrough only when replay-safe), how to disable each layer, and a "circuit breaker vs dashboard provider health" section. `CACHE_REFRESH_INTERVAL` enters the user-facing configuration reference (it was only an `.env.template` comment).

## User-visible impact

- A provider outage now shows as Degraded with its models intact, is re-probed every minute by default, and recovers into rotation within ~a minute of coming back.
- Load-balanced redirects keep routing around down providers exactly as before.
- Streaming-heavy workloads can no longer trip a healthy provider's breaker through client disconnects.
- New env vars: `PROVIDER_RECHECK_INTERVAL` (60); new metric: `gomodel_circuit_breaker_state`.

## Testing

- 12 new regression tests (breaker wedge + cancellation, carry-forward/recovery, stale LB skip, bare-ID duplicate preference, sweep starvation, recheck loop, gauge, hooks plumbing, dashboard timestamp). The two breaker bugs were reproduced as failing tests before the fix; the starvation test was verified to fail against the old sequential sweep.
- Full `go test ./...` green, `-race` clean on all touched packages, all 400 dashboard JS tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added `PROVIDER_RECHECK_INTERVAL` cache setting (default 60s) to re-probe only failed providers.
  * Dashboard health now distinguishes degraded providers and updates “Last checked” using the most recent check.
  * Added Prometheus `gomodel_circuit_breaker_state` gauge for circuit-breaker visibility.
* **Bug Fixes**
  * Failed provider refreshes now keep serving last-known models while marking inventory stale, preventing redirects/load balancing from selecting degraded targets.
  * Circuit breaker now correctly releases half-open probe slots after cancellations and build errors.
* **Documentation**
  * Updated configuration, resilience, and Prometheus metrics docs to reflect the new behavior and settings.


SantiagoDePolonia:
@coderabbitai continue

SantiagoDePolonia:
**Re: P1 "Carry-forward is skipped" (registry_init.go early return)** — addressed in 9b663e7c, though not by marking stale on the total-failure sweep, which would make things worse:

- If every provider fails, there is no healthy target left. Marking all inventory stale would empty `supportedTargets` for every redirect, and resolution falls through to the literal alias name → clients get **404 "model not found"** instead of the provider-level 502/503 they get by routing through. When nothing is healthy, routing to a down provider is the more honest degradation (and lets breaker fail-fast + failover rules work).
- A whole-sweep failure can also be control-plane-only (gateway can't reach any `/models` endpoint while inference works). Marking everything stale would break all aliased traffic on a false positive. This early return predates the stale mechanism for exactly this reason; it's now documented in-code and pinned by `TestInitialize_TotalRefreshFailureKeepsRouting`.

The real gap behind the finding was the recovery window: after a total outage, a provider recovering via the 60s recheck left its still-down peers unmarked (fresh) until the next full sweep — LB kept routing to them for up to an hour. Fixed: failed per-provider probes (recheck loop, request-time refresh, availability checks) now mark the provider's carried inventory stale **as soon as at least one healthy alternative exists**, retiring it from load balancing within ~one recheck interval. Covered by `TestRefreshProviderModels_TotalOutageRecoveryRetiresStillDownPeer`, `..._FailureMarksStaleWhenAlternativeHealthy`, and `..._AvailabilityFailureMarksStale`.

SantiagoDePolonia:
@coderabbitai continue

PR #483, opened 2026-07-05
fix(config): honor the http: YAML block for upstream client timeouts

SantiagoDePolonia:
## Summary

Closing out the architecture review's config escape-hatch item — with a twist found during verification.

### The real defect (fixed)
`config.HTTPConfig` (`http:` YAML block) was parsed, defaulted, and documented in `config.example.yaml` — but **no code ever read it**. Setting `http.timeout` / `http.response_header_timeout` in YAML was silently ignored; only the `HTTP_TIMEOUT` / `HTTP_RESPONSE_HEADER_TIMEOUT` env vars worked, via `httpclient`'s direct env read.

Fix: `httpclient.SetConfiguredTimeouts` installs the YAML values at app startup, before any provider constructs a transport. Precedence (highest first): env var → YAML `http:` block → built-in 600s default, matching the project-wide env-over-YAML convention. Pinned by a test covering all three tiers plus clearing.

### Findings re-verified, not blindly applied
- The review claimed `OPENROUTER_SITE_URL`/`OPENROUTER_APP_NAME` were "undocumented hidden config" — **false positive**: both are in `.env.template` and the provider docs. The review doc is corrected in place.
- The remaining per-provider construction-time env reads (openrouter, opencodego) are kept deliberately — that's the established quirk-knob convention, and plumbing them through `ProviderConfig` would add ceremony for no behavior gain.
- The anthropic request-time `ANTHROPIC_DEFAULT_MAX_TOKENS` read keeps its warn-once + safe-fallback behavior; threading it through provider construction wasn't worth the signature churn. Recorded in the review doc.
- CLAUDE.md's provider list gains the missing mentions (`OPENROUTER_SITE_URL`/`APP_NAME`, `ANTHROPIC_DEFAULT_MAX_TOKENS`) and the HTTP bullet now notes the YAML block.

## User-visible impact

YAML-configured HTTP timeouts now actually apply. Anyone relying on the documented `http:` block gets the behavior the docs always promised; env-var users see no change.

## Testing

Full `go build ./...` and `go test ./...`; new precedence test in `internal/httpclient`; `golangci-lint` 0 issues; pre-commit `make test-race` on the commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP client timeout settings from the app `http:` configuration are now applied during startup.
  * Environment variables still override configuration, and cleared/non-positive values fall back to built-in defaults.
* **Documentation**
  * Updated HTTP timeout and provider environment-variable documentation to clarify precedence and added/clarified provider attribution and token-limit settings.
* **Tests**
  * Added tests covering timeout precedence across built-in defaults, configured values, and environment overrides.


SantiagoDePolonia:
@copilot Resolve conflicts in CLAUDE.md - merging main

PR #487, opened 2026-07-05
feat(headers, provider): custom header overrides and Moonshot Kimi provider

weselben:
This PR adds per-provider HTTP header overrides and a new Moonshot Kimi provider.

## Custom upstream / passthrough headers

A new `HeaderOverridesConfig` engine lets operators control how headers flow to upstream providers.

- `custom_upstream_headers`: static `map[string]string` headers attached to every request for a provider.
- `passthrough_user_headers`: forwards selected client headers upstream.
- `passthrough_user_headers_skip` + `passthrough_user_headers_skip_mode`: configure a skip or allow list (default: `skip`).
- A hard-coded security floor always strips credential headers (`Authorization`, `Cookie`, `X-Api-Key`, etc.), the internal `X-GoModel-User-Path` header, and any configured `userPathAlias`.

Example:

```yaml
providers:
  kimi:
    api_key: ${KIMI_API_KEY}
    custom_upstream_headers:
      X-Title: "MyApp"
    passthrough_user_headers: true
    passthrough_user_headers_skip:
      - X-Trace-Id
    passthrough_user_headers_skip_mode: skip
```

The feature is wired into the factory and OpenAI-compatible adapter, and is applied by DeepSeek, Groq, Ollama, vLLM, xAI, and the new Kimi provider. The server middleware that captures passthrough headers is gated so the default hot path is unchanged when no provider opts in.

## Moonshot Kimi provider

- New `internal/providers/kimi` package wrapping the OpenAI-compatible client.
- Default base URL: `https://api.kimi.com/coding/v1`.
- Registered in the factory, config examples, Helm values, and provider overview docs.
- Supports `custom_upstream_headers` and `passthrough_user_headers` via the shared override engine.
- Unit tests, contract tests, golden files, and `recordapi` tooling for generating Kimi test fixtures.

## Known Limitations

- Header overrides are not yet implemented for OpenRouter or native SDK providers (Anthropic, Gemini, Vertex, Bedrock, etc.).
- The passthrough path is only wired for OpenAI-compatible providers in this PR.

## Commits

- dbd8841 docs: add custom headers feature guide and Kimi provider docs
- 50d0cc1 feat(headers): add custom upstream and passthrough header overrides
- 3efecb3 tests(headers): add unit and provider tests for header overrides
- c3d5214 feat(provider): add Moonshot Kimi provider
- 0503eb7 tests(provider): add Kimi contract tests, golden fixtures and recordapi support

## Future scope

The following native / non-OpenAI-compatible providers are not covered by the header override feature in this PR, but could be wired in later if the same configuration surface is desired:

- Anthropic
- Google Gemini / Vertex AI
- Amazon Bedrock
- Azure OpenAI
- MiniMax
- Xiaomi MiMo
- OpenCode Go
- Z.ai

Each provider uses its own SDK or native request path, so adding support requires a per-provider audit of how headers are constructed before forwarding. OpenRouter was intentionally left out of this PR because it already sets its own attribution headers (`HTTP-Referer`, `X-OpenRouter-Title`, etc.) and changing that behavior was out of scope. OpenAI-compatible providers (including Kimi, OpenAI, DeepSeek, Groq, Ollama, vLLM, xAI, Fireworks AI, and Oracle) already inherit the feature through the shared adapter.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Kimi provider, including chat, streaming, models, and embeddings requests.
  * Added configurable upstream header handling, including custom headers and optional passthrough of selected user headers.
  * Expanded supported provider configuration to include additional Kimi settings and example setups.

* **Bug Fixes**
  * Improved header filtering so sensitive or internal headers are not forwarded upstream.
  * Enabled passthrough header capture only when a provider requires it.


SantiagoDePolonia:
_After a discussion with Fable 5 — summarized by AI._

This needs a round of changes before it can merge. Findings below, ordered by severity.

## Blockers

**1. The branch doesn't compile and CI is red.**
The new tests in `internal/app/app_test.go` were inserted *inside* the body of `staticRewriter.Rewrite` — its closing brace is missing, so the whole package fails to parse (`app_test.go:649: expected '(', found TestAnyProviderHasPassthroughUserHeaders_True`). Both the **Unit Tests** and **lint** checks are failing on this. Additionally, five new files are missing trailing newlines (`config/providers_test.go`, `internal/providers/kimi/kimi.go`, `internal/providers/kimi/kimi_test.go`, `internal/server/passthrough_headers.go`, `tests/contract/kimi_test.go`) and several struct blocks are not gofmt-aligned (`config/providers.go`, `internal/providers/factory.go`, `internal/providers/openai/compatible_provider.go`). Please run `gofmt` and `go test ./...` locally before the next push.

**2. The passthrough engine bypasses the tagging `do_not_pass` guarantee.**
`PassthroughHeaderCapture` runs before the tagging middleware and snapshots all non-credential headers into the context; `ApplyHeaderOverrides` never consults the tagging strip set. Consequences:
- On **translated routes**, a tagging header the operator explicitly marked `do_not_pass` is now forwarded upstream whenever a provider enables `passthrough_user_headers` (these routes previously never forwarded client headers).
- On **`/p/` passthrough routes** it's worse: in `llmclient.buildRequest`, the header setter (which now applies the context capture) runs *before* the per-request headers built by `passthrough_support.go`, and only keys present in `req.Headers` get deleted. Headers that `passthrough_support` deliberately stripped (do-not-pass headers, hop-by-hop `Connection`-listed headers) are re-added by the override engine.

The engine needs to respect the tagging strip set (`core.TaggingStripHeadersFromContext`) and the hop-by-hop set.

**3. Forwarding `Accept-Encoding` breaks response decoding.**
Go's HTTP transport only transparently decompresses gzip when *it* set `Accept-Encoding`. The default passthrough mode (`skip` + empty list) forwards the client's `Accept-Encoding: gzip, deflate, br, …` (which nearly every SDK sends), and `config/config.example.yaml` even recommends `Accept-Encoding: "gzip, deflate, br"` as a static custom header. Either way the upstream responds compressed and `llmclient` feeds compressed bytes to `json.Unmarshal` — every translated call (chat *and* embeddings) fails. The replay tests can't catch this because the mock never compresses. Please add a hard blocklist for transport/content-negotiation headers (`Accept-Encoding`, `Connection`, `TE`, `Keep-Alive`, `Transfer-Encoding`, `Content-Length`, `Host`) and remove `Accept-Encoding`/`Content-Type` from the example config.

## Design changes

**4. The feature applies to far more providers than the PR and docs claim.**
Because `NewCompatibleProvider` auto-injects `opts.HeaderOverrides`, every compatible-adapter provider inherits it: OpenRouter, Azure, MiniMax, Xiaomi, Z.ai, Bailian, Oracle, OpenCode-Go, Fireworks — not just the six listed. `docs/features/custom-headers.mdx` says OpenRouter "ignores" the config; in reality passthrough can clobber OpenRouter's attribution headers via the delete-then-add rule — exactly what the PR description says it wanted to avoid. Either gate the auto-inject to the intended providers or (better) embrace the wider coverage and fix the docs, with a test that OpenRouter attribution headers survive.

**5. `custom_upstream_headers` and `passthrough_user_headers` should compose, not be mutually exclusive.**
With passthrough enabled, static headers are silently dropped (debug-level log only). The PR description's own Kimi example sets *both* — under the current semantics the static identity headers it demonstrates would never be sent. Suggested order: apply statics first, then passthrough (client values winning per header), or the reverse — but pick one and document it.

**6. Default-open forwarding is a risky default.**
`skip` mode with an empty list forwards *all* non-credential client headers (`traceparent`, `X-Forwarded-For`, internal tracing/baggage headers…). Consider making an empty configuration forward nothing (require an explicit allowlist), or at minimum add a prominent doc warning.

**7. `SetUserPathHeader` is called after all providers are constructed.**
In `internal/app/app.go`, `cfg.Factory.SetUserPathHeader(...)` runs after `providers.Init` has already created every provider, so `opts.UserPathHeader` is always `""` at provider level — the per-provider alias blocking is dead code (currently masked by the server-side capture filter). Set it before `Init`, or pass it through the provider config.

**8. The claimed passthrough defaults don't match the code.**
The updated comment in `config/server.go` and CLAUDE.md claim the default `ENABLED_PASSTHROUGH_PROVIDERS` now includes `bailian` and `kimi`, but `defaultEnabledPassthroughProviders` in `internal/server/passthrough_support.go` is unchanged — `/p/kimi/...` doesn't work out of the box. Update the actual default or the docs, not just the comments.

## Kimi provider: naming, default URL, and docs accuracy

We looked at Kimi's official documentation. There are two distinct products:

- **Moonshot open platform** — `https://api.moonshot.ai/v1`, pay-as-you-go, real model catalog (`kimi-k2-thinking`, `kimi-k2-turbo-preview`, …), documented endpoints `/v1/chat/completions` + `/v1/files*`, **no `/v1/embeddings`**.
- **Kimi Code** (formerly Kimi for Coding) — `https://api.kimi.com/coding/v1`, a kimi.com membership benefit with its own API-key realm (keys are not interchangeable with platform keys), exactly one documented model (`kimi-for-coding`, a stable alias), and quota semantics (weekly refresh + rolling 5-hour rate window) instead of per-token billing.

The wire dialect is the same OpenAI-compatible format, so a separate provider *type* isn't justified — but defaulting a type named `kimi` to the membership-gated coding endpoint will surprise anyone holding a platform key. Please either:

- **(preferred)** default the `kimi` type to `https://api.moonshot.ai/v1` and document Kimi Code as a configured instance via the existing env-suffix convention (`KIMI_CODE_BASE_URL=https://api.kimi.com/coding/v1`, `KIMI_CODE_API_KEY`, `KIMI_CODE_MODELS=kimi-for-coding` → provider `kimi-code`), or
- if this provider is deliberately Kimi Code only, name the type `kimi-code` to match the official product name.

Related docs/tooling fixes:

- **Remove the spoofed client-identity headers** (`User-Agent: KimiCLI/0.76`, `X-Title: Kimi CLI`, `X-Stainless-*`, `Http-Referer`) from `cmd/recordapi/main.go` and `config/config.example.yaml`. Kimi officially supports third-party tools with nothing but the base URL and API key — these headers are unnecessary, and shipping impersonation examples isn't something this repo should endorse.
- **Remove the "ZooCode identity headers" section** from `docs/providers/kimi.mdx` — we could find no such feature in Kimi's documentation (`X-ZooCode-User`/`X-ZooCode-Plan`/`X-ZooCode-Team` appear to be fabricated).
- **Embeddings**: `bge_m3_embed` works today but is undocumented upstream — mark it as such ("may break without notice") and drop the nonexistent `kimi-embedding` example (the docs currently use both IDs inconsistently).
- **Pricing/limits**: note that Kimi Code has no per-token pricing, so usage-cost tracking will show zero and the `cost` load-balancing strategy can't price it; given the rolling 5-hour quota window, recommend a conservative per-provider `resilience.retry` override so 429 retries don't burn quota.

## Smaller cleanups

- `InitResult.ResolvedProviders` duplicates the full credential-bearing provider map just so `app.go` can compute one boolean — expose a bool (or compute it inside `Init`) instead.
- `internal/providers/vllm/vllm.go`: the new `headerOverrides`/`userPathAlias` struct fields appear unused (the closures already capture them).
- `docs/features/custom-headers.mdx`: the allowlist example uses an `anthropic:` provider block, which the same page's Limitations section says ignores the feature.

## Suggestion: split the PR

With the spoofed headers removed, the Kimi provider doesn't depend on the header-override feature at all. Splitting this into (a) the Kimi provider and (b) the header-override engine would make both much easier to land — (a) is close to ready once naming/docs are settled; (b) needs the security and composition rework above.

PR #494, opened 2026-07-06
feat(virtualmodels): validate target providers at startup

### LINKED ISSUE #464: Improved virtual model validation
GoModel comes up and says everything is alright, but there is a typo in the `general` virtual model targe provider that is not caught. (an extra 'n'). I would see validating the target providers against known providers, before any API calls are made for checking if it is up.

OpenCode reports "unsupported model: general"

```yaml
server:
  port: "9999"

virtual_models:
- source: coder-moe
  targets:
  - provider: plutonium-coder-moe
    model: Qwen/Qwen3.6-35B-A3B-FP8
- source: coder-dense
  targets:
  - provider: plutonium-coder-dense
    model: Qwen/Qwen3.6-27B-FP8
- source: general
  targets:
  - provider: uraninum-general
    model: nvidia/Gemma-4-26B-A4B-NVFP4
- source: minillm
  targets:
  - provider: uranium-minillm
    model: Qwen/Qwen3-1.7B
- source: reranker
  targets:
  - provider: uranium-reranker
    model: Qwen/Qwen3-Reranker-0.6B
- source: embedding
  targets:
  - provider: uranium-embedding
    model: google/embeddinggemma-300m

models:
  enabled_by_default: true # env: MODELS_ENABLED_BY_DEFAULT; when false, models stay unavailable until an access override allows one or more user paths
  configured_provider_models_mode: "fallback" # env: CONFIGURED_PROVIDER_MODELS_MODE; "fallback" uses configured lists only when upstream /models is unavailable/empty, "allowlist" exposes only configured models and skips upstream /models for configured lists

providers:
  plutonium-coder-moe:
    type: vllm
    base_url: "http://plutonium-coder-moe:8000/v1"
  plutonium-coder-dense:
    type: vllm
    base_url: "http://plutonium-coder-dense:8000/v1"
  uranium-general:
    type: vllm
    base_url: "http://uranium-general:8000/v1"
  uranium-minillm:
    type: vllm
    base_url: "http://uranium-minillm:8000/v1"
  uranium-reranker:
    type: vllm
    base_url: "http://uranium-reranker:8000/v1"
  uranium-embedding:
    type: vllm
    base_url: "http://uranium-embedding:8000/v1"

logging:
  enabled: true
  log_bodies: true # WARNING: may contain sensitive data
  log_headers: true
  buffer_size: 1000
  flush_interval: 5 # seconds
  retention_days: 30 # 0 = keep forever
  only_model_interactions: true

metrics:
  enabled: true
  endpoint: "/metrics"

http:
  timeout: 1800 # seconds (30 minutes)
  response_header_timeout: 1800

workflows:
  refresh_interval: 1m

# Global resilience settings (applied to all providers by default)
# Individual providers can override any of these values.
resilience:
  retry:
    max_retries: 3
    initial_backoff: 6s
    max_backoff: 30s
    backoff_factor: 2.0
    jitter_factor: 0.1

```

SantiagoDePolonia:
Closes #464

## Problem

A virtual model target naming a provider that doesn't exist — e.g. the issue's `uraninum-general` typo for `uranium-general` — booted fine and only surfaced per-request as an unhelpful "unsupported model: general". Startup said everything was alright.

## What changed

Startup now validates every **explicitly-named** target provider of declarative (`virtual_models:` / `VIRTUAL_MODELS`) redirects against static provider configuration, before any API call, in three tiers:

- **Registered provider** → passes. Provider names register synchronously at startup, so this works even on a cold model catalog.
- **Declared under `providers:` but unregistered** (typically credentials unset in this environment) → logs a warning; the target stays unavailable until the credentials resolve. A `config.yaml` shared across environments with per-env API keys still boots everywhere.
- **Unknown everywhere** (a typo) → aborts startup with a validation error listing the registered provider names:

  ```
  failed to initialize virtual models: load virtual model "general": unknown target provider "uraninum-general" (registered providers: plutonium-coder-moe, uranium-general, uranium-minillm)
  ```

Deliberate scope limits:

- Only the explicit `provider:` field is checked. Provider-less targets (including the `target:` shorthand) are skipped because slash-shaped model IDs (e.g. HF org prefixes like `Qwen/Qwen3-1.7B`) are not provider references.
- Target **model** availability remains a resolve-time concern — the catalog loads asynchronously and must not gate startup (preserves the cold-catalog fix; its regression tests still pass).

The admin write path reuses the same check, so `PUT /admin/virtual-models` with a misspelled provider now returns a precise `unknown target provider "..."` 400 instead of the vaguer "target model not found".

Related discoverability fix: providers dropped during credential resolution were silently removed; `providers.Init` now logs one line naming them (`configured providers skipped: credentials or base_url did not resolve`).

## User-visible impact

- A previously-booting config with a typo'd virtual-model target provider now **fails startup** with an actionable error. This is intentional fail-fast behavior for a declaration that can never work; configs with declared-but-uncredentialed providers keep booting (warning only).
- Better admin API error message for misspelled providers.
- New startup log line listing credential-skipped providers.

## Testing

- Unit: three-tier table tests (typo aborts, declared-keyless passes with warning, registered passes, slash-model target never treated as a provider, typo among valid targets still caught) plus admin-upsert rejection; existing cold-catalog and structural-validation tests unchanged and green.
- E2E: `tests/e2e/test-iac-virtualmodels.sh` extended with I12 (typo aborts startup) and I13/b/c (declared-keyless boots with warning, redirect 404s, skipped-providers log present) — 19/19 PASS against real providers.
- Manually reproduced the issue's exact config: typo aborts, corrected config boots healthy.
- `make test-race`, `go vet`, `gofmt`, lint, hot-path perf guards all green.

## Docs

`docs/features/virtual-models.mdx` and the CLAUDE.md config reference document the new semantics.


SantiagoDePolonia:
@coderabbitai continue

PR #511, opened 2026-07-08
fix(config): reject unknown config keys and log provider origins

### LINKED ISSUE #509: Validation with ENV and config
I have an OpenCode token I pass as an ENV var and that automatically provisions the provider upon detection.

I also have a config file I pass with my vLLM providers. I generated some bad yaml and the issue was hidden from me until I removed the OpenCode provider, at which point the server started crashing (no providers configured) and I was able to narrow down why my local models were not showing up.

I'm not sure what to do here validation wise, but it would be helpful to have a boot log line that shows how many providers come from config vs env vars. I noticed there was only one, so I disabled it, and then got my crash.

---

Bad (providers) yaml

```yaml
$ cat tmp.yml 
server:
  port: "9999"

virtual_models:
- source: geryon-9b
  targets:
  - provider: uranium-geryon-9b
    model: geryon-9b
- source: gemma-embedding
  targets:
  - provider: uranium-gemma-embedding
    model: gemma-embedding
- source: qwen-reranker
  targets:
  - provider: uranium-qwen-reranker
    model: qwen-reranker
- source: qwen-dflash
  targets:
  - provider: plutonium-qwen-dflash
    model: qwen-dflash

providers:
uranium-geryon-9b:
  type: vllm
  base_url: "http://uranium-geryon-9b:8000/v1"
uranium-gemma-embedding:
  type: vllm
  base_url: "http://uranium-gemma-embedding:8000/v1"
uranium-qwen-reranker:
  type: vllm
  base_url: "http://uranium-qwen-reranker:8000/v1"
plutonium-qwen-dflash:
  type: vllm
  base_url: "http://plutonium-qwen-dflash:8000/v1"

...

```

SantiagoDePolonia:
Closes #509

## Problem

A misindented `providers:` section is valid YAML — just not the YAML you meant:

```yaml
providers:
uranium-geryon-9b:        # column zero: a new top-level key, not a provider
  type: vllm
  base_url: "http://uranium-geryon-9b:8000/v1"
```

It parses as `providers: null` plus four unrelated top-level keys, which `yaml.Unmarshal` discarded silently. The gateway booted with none of the operator's providers and no error. In #509 the fault stayed hidden because an env-provisioned OpenCode provider kept the registry non-empty; removing it turned the silent misconfiguration into `no providers were successfully registered`, which points nowhere near the real cause.

## User-visible impact

**Unknown keys in declarative config now fail startup by default**, naming the file and every offending line:

```
failed to load config: failed to parse config.yaml: line 23: field uranium-geryon-9b not found; line 26: field uranium-gemma-embedding not found; line 29: field uranium-qwen-reranker not found; line 32: field plutonium-qwen-dflash not found
```

This also catches ordinary typos (`prot:` for `port:`, `bass_url:` for `base_url:`).

**A boot log line reports where providers came from**, as requested in #509:

```json
{"msg":"config file loaded","path":"config.yaml","providers":4}
{"msg":"providers resolved","total":5,"from_config_file":4,"from_env":1,
 "config_file_providers":["plutonium-qwen-dflash","uranium-geryon-9b"],
 "env_providers":["opencode-go"]}
```

`from_config_file: 0` while a file is loaded means the file contributed nothing. There is also `no config file found` listing the searched paths.

## `CONFIG_STRICT=false` escape hatch

Strict is the right default: a dropped `providers`, `rate_limits`, `budgets`, or `guardrails` entry silently changes routing, cost, or security — a cost cap that never applies looks exactly like a cost cap that does. Strict's worst case is a boot failure with a precise message; liberal's worst case is a gateway running for weeks with a control silently absent.

But it blocks one legitimate case: rolling a binary back under a config file written for a newer version, where an unknown key is expected rather than a typo. `CONFIG_STRICT=false` downgrades unknown keys to warnings and boots without them:

```text
WARN  CONFIG_STRICT=false: unknown config keys are ignored with a warning instead of aborting startup
WARN  unknown config key ignored; it has no effect path=config.yaml line=23 field=uranium-geryon-9b
INFO  providers resolved total=1 from_config_file=0 from_env=1
```

Three properties make the lax mode meaningfully better than the old silent behaviour:

- **Unknown keys are always detected.** Lax mode still decodes with `KnownFields(true)` and inspects the resulting `TypeError`, so it *names* each ignored key rather than dropping it in silence the way `yaml.Unmarshal` did.
- **The flag relaxes which keys are accepted, never whether a value makes sense.** A malformed value (`port: [9999, 8080]`) or a YAML syntax error stays fatal in both modes, and a file mixing an unknown key with a malformed value still fails.
- **Entity-level validation is unaffected.** A virtual model whose target names a provider that does not exist still fails startup even when the key that would have declared that provider was only warned about.

## Why the env layer changed too

Per the documented priority (env > yaml > defaults), env entries override YAML entries key by key. Leaving env JSON lax while YAML is strict recreates the same trap one layer up: a typo in `VIRTUAL_MODELS`, `SET_RATE_LIMIT_*`, or `SET_BUDGET_*` would silently win over a correct YAML entry. All three now decode through one shared helper that honours `CONFIG_STRICT` and downgrades only `encoding/json`'s unknown-field error.

Two things fall out of that:

- **A latent bug.** `BudgetLimitConfig` had no json tags, unlike its `RateLimitRuleConfig` neighbour which added them for exactly this env form. `SET_BUDGET_X='[{"period_seconds":7200,"amount":5}]'` silently dropped `period_seconds`, producing a limit with no window. Verified against the old code, then fixed by adding the tags.
- **`failover.overrides`.** A removed key that old config files still carry, and that existing tests deliberately assert still loads. Strict parsing would have broken that contract, so it is modelled as an explicitly ignored field — upgrading it from *silently* ignored to a logged deprecation pointing at `disabled_models`.

## Also fixed

A config path that exists but cannot be read — most often a directory bind-mounted where a file was expected, a common Docker mistake — was indistinguishable from a missing file and silently fell back to defaults. It is now reported: `failed to read config.yaml: read config.yaml: is a directory`.

## Breaking change

With the default `CONFIG_STRICT=true`, configs carrying stray keys will not boot. The error names the key and the line, and `CONFIG_STRICT=false` is the documented out. Top-level YAML anchors (`x-defaults: &d`) are also rejected under the default — nothing in the repo or docs uses them, and anchors defined inside `providers:` still work.

## Verification

Full suite passes; `golangci-lint` clean; all pre-commit hooks (including `make test-race`) pass. Beyond unit tests, the built binary was driven through every path against the exact config from #509:

| Scenario | Result |
| --- | --- |
| Reporter's file, default | fails, naming all four lines |
| Reporter's file, `CONFIG_STRICT=false` | 4 warnings, boots, `from_config_file: 0` |
| Reporter's file, indentation fixed | boots, 4 vLLM providers, virtual models validate |
| `CONFIG_STRICT=false` + `port: [9999, 8080]` | still fatal |
| `CONFIG_STRICT=maybe` | rejected as non-boolean |
| Directory mounted as `config.yaml` | fails with a read error |
| Comments-only file | boots |

New tests cover misindentation, nested and per-provider typos, empty and comments-only files, the `...` document-end marker the reporter's file uses, the unreadable-path case, both strict and lax modes in YAML and env JSON, type errors staying fatal when lax, the budget json-tag fix, and a regression test that `config.example.yaml` itself parses strictly — otherwise anyone copying it could not boot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unknown or misspelled fields in `config.yaml` and structured JSON-style environment settings are rejected by default (malformed values remain fatal); multi-document YAML is rejected.
  * Strict JSON env parsing now also rejects trailing/unparsed data.
  * Mounted config paths that exist but can’t be read now fail startup instead of falling back.
* **New Features**
  * `CONFIG_STRICT` (default: `true`) controls whether unknown keys are rejected or downgraded to warnings.
  * Startup logs report provider resolution counts and origins (from file vs. environment).
* **Documentation**
  * Updated strictness/overlay gotchas and startup validation behavior, including strict handling of misindented `providers`.  
* **Chores**
  * Bumped Go toolchain/build images.


SantiagoDePolonia:
Addressed both bot findings.

**Trailing JSON (P1 / Major — @greptile-apps, @coderabbitai).** Confirmed and fixed. Swapping `json.Unmarshal` for a `json.Decoder` to get `DisallowUnknownFields` silently regressed trailing-data rejection:

```
Unmarshal:  err=invalid character 'T' after top-level value
Decode:     err=<nil>  v=[{Period:daily}]     ← accepted `[{...}] THIS IS GARBAGE`
```

`decodeStrictJSON` now requires `io.EOF` after the decoded value. Tests cover a garbage suffix, a concatenated second JSON value (Greptile's exact repro), and a newline-separated one — each in both strict and lax mode, since trailing data is a structural fault rather than an unknown key and stays fatal under `CONFIG_STRICT=false`. Verified that trailing whitespace is still accepted.

**The same hole existed in the YAML layer**, pre-dating this branch and unflagged: `yaml.Unmarshal` and `yaml.Decoder` both read exactly one document, so every key after a `---` separator was applied nowhere. Now rejected, also fatal in both modes. The `...` document-end marker still parses.

**Docs (Minor — @coderabbitai).** Correct: the prose counted keys the sample did not contain. The sample now carries two providers, and the quoted error string is the binary's verbatim output rather than my paraphrase.

Verified with unit, contract, e2e, and Docker-backed integration suites, `golangci-lint` across all build tags, and by driving the real binary through each path.

SantiagoDePolonia:
## Note: this PR now also bumps Go to 1.26.5

Unrelated to the config changes, folded in at the author's request because it blocks CI.

`govulncheck` began reporting **GO-2026-5856** (Encrypted Client Hello privacy leak in `crypto/tls`, fixed in go1.26.5). It is **not a regression from this branch** — I reproduced it on pristine `origin/main` at `e0eb0eef` with no changes applied:

```
Fixed in: crypto/tls@go1.26.5
Your code is affected by 1 vulnerability from the Go standard library.
```

The reachable traces are the existing dashboard, redis, and mongo TLS paths. `main` is red right now for the same reason.

Bumped every pin, so the **shipped binary** is patched rather than just CI:

| Location | Before | After |
| --- | --- | --- |
| `.github/workflows/test.yml` | `GO_VERSION: "1.26.4"` | `"1.26.5"` |
| `.github/workflows/release.yml` | `go-version: 1.26.4` | `1.26.5` |
| `Dockerfile` | `golang:1.26.4-alpine3.23` | `golang:1.26.5-alpine3.23` |
| `go.mod` | `go 1.26.4` | `go 1.26.5` |

Verified locally on go1.26.5: `govulncheck` reports **no vulnerabilities** both with and without the `swagger` tag; unit, contract, e2e, and Docker-backed integration suites pass; `golangci-lint` is clean across all build tags; and the release image builds and reports `gomodel dev (... go1.26.5)`.

If you would rather keep this PR focused, the last commit (`91db16b8`) cherry-picks cleanly onto `main` on its own.

PR #521, opened 2026-07-12
feat(admin): reflect real-traffic request health in provider status

SantiagoDePolonia:
## Problem

A provider whose upstream is failing real requests (e.g. OpenCode Zen answering `400 "Error from provider"` for one model) still showed **Healthy** in the dashboard, because provider status was derived only from model-list discovery. Circuit-breaker state was invisible outside the Prometheus gauge.

## What this does

**Request-health tracking** (`internal/providers/health`)
- New `Tracker` fed purely by existing `llmclient` observability hooks (composed with the Prometheus hooks via new `llmclient.JoinHooks` / `factory.AddHooks`) — no new outbound traffic, no hot-path changes.
- Keeps a 10-minute sliding window of per-provider/per-model request and error counts, the last error per model, and the last observed circuit-breaker state. Memory-bounded (100 models/provider, 500 events/model). Background discovery/availability probes (`model: "unknown"`) are excluded from traffic counts.
- A model is **flagged** when it has ≥3 windowed errors making up ≥50% of its requests, so a single malformed user request never flags a model.

**Merged provider status** (`/admin/providers/status`)
- Each provider row now carries `request_health`; real-traffic signals fold into the status pill, only ever worsening it:
  - breaker **open** → `Circuit Open` (unhealthy)
  - breaker **half-open** → `Recovering` (degraded)
  - flagged model(s) on an otherwise healthy provider → `Degraded` with "recent requests are failing for: …" and the model's latest upstream error surfaced as `last_error`
- Discovery-based classification is unchanged otherwise.

**Dashboard cards**
- Expanded details show Recent Requests (windowed totals), a live Breaker State chip, and per-model traffic rows with flagged models highlighted and the last error on hover.
- Cards are now collapsible one-by-one via a full-width bottom toggle strip (rotating chevron, hover highlight, keyboard accessible); per-card state persists in localStorage and the section-wide switch remains as a master control. Collapsed cards no longer stretch to match expanded neighbours.
- The status pill shows the classification reason and last error on hover.

## Notes

- Tracking is in-memory per instance and resets on restart (same as rate-limit counters).
- 4xx responses still never trip the circuit breaker (a bad request must not black-hole a provider); they surface through the flagged-model path instead.
- Verified live: bad OpenAI key + stopped Ollama → pills flip to Degraded/Unhealthy with per-model detail; background probes generate no phantom traffic rows.

## Summary by CodeRabbit

* **New Features**
  * Added live request-health (last 10 minutes) and circuit-breaker state to provider status, including per-model recent traffic/error details and flagged indicators.
  * Updated the admin providers status response to include request-health data when available.
* **UI / Styling**
  * Improved provider cards with per-provider expand/collapse (persisted across sessions) and a richer status pill tooltip; details now show recent requests, breaker state, and a models list.
* **Documentation**
  * Refreshed resilience and provider overview docs to explain combined dashboard health behavior (discovery + request health).
* **Tests**
  * Expanded coverage for request-health integration, dashboard helpers, and UI toggle persistence.


SantiagoDePolonia:
Addressed the review findings in e17ed88b:

- **Client cancellations flag providers** (greptile P1) — fixed. `Tracker.Record` now treats `context.Canceled` as neutral, matching the circuit breaker's semantics; client deadlines (`DeadlineExceeded`) still count as failures.
- **Trimmed names drop health** (greptile P2) — fixed. `requestHealthFor` matches snapshot keys against the trimmed row name.
- **Snapshot cap can hide the true most-recent error** (coderabbit) — fixed. `ProviderHealth` now carries a provider-level `last_error`/`last_error_model` computed across all tracked models before the 25-row cap; `latestModelError` uses it.
- **UTF-8 mid-rune truncation** (coderabbit) — fixed with a rune-boundary cut.
- **Hardcoded "unknown" sentinel** (coderabbit) — now exported as `llmclient.UnknownModel` and shared.
- **Missing breaker-vs-flagged precedence test** (coderabbit) — added.
- **Streaming failures look successful** (greptile P1) — acknowledged as a known limitation, now documented in the package comment. Hooks fire at stream establishment by design (the Prometheus hooks share the same blind spot); observing mid-stream failures would require changes to the llmclient hook lifecycle, which is out of scope for this PR.

PR #522, opened 2026-07-13
chore(build): rename module path to github.com/enterpilot/gomodel

SantiagoDePolonia:
## Summary

Renames the Go module from `gomodel` to `github.com/enterpilot/gomodel` (canonical, fetchable path).

## Why

- Makes the project fetchable via the Go module proxy and indexable on **pkg.go.dev** (currently 404s because `module gomodel` is not a resolvable path).
- Enables `go install github.com/enterpilot/gomodel/cmd/gomodel@latest`.
- Required for the awesome-go submission: "pkg.go.dev reachable" is a blocking CI check there.

## Changes

- `go.mod`: `module gomodel` → `module github.com/enterpilot/gomodel`
- Mechanical rewrite of all `"gomodel/..."` imports across 497 Go files
- `-X gomodel/internal/version.*` ldflags updated in `Makefile`, `Dockerfile`, `.goreleaser.yaml`
- Bare `"gomodel"` strings (default Mongo DB name, CLI product name) intentionally untouched

## Verification

- `go build ./...` clean
- `go test ./...` all green
- `go vet` shows only the three pre-existing `ContentSchema` json-tag warnings (present before the rename)
- No `"gomodel/"` references remain

After merging, tag a release (e.g. v0.1.52) so pkg.go.dev indexes the new path.


SantiagoDePolonia:
Added a `debug.ReadBuildInfo()` fallback in `internal/version`: `go install github.com/enterpilot/gomodel/cmd/gomodel@latest` builds now report the module version (and VCS commit/date when stamped) instead of `dev (commit: none, built: unknown)`. ldflags-stamped builds (Makefile/Docker/GoReleaser) are unchanged and still take precedence.

PR #541, opened 2026-07-17
feat(config): accept GOMODEL_-prefixed env vars, deprecate bare names

SantiagoDePolonia:
## What

GoModel-defined environment variables are now canonically spelled `GOMODEL_<NAME>`. The unprefixed spelling still resolves, so **no existing deployment breaks** — but reading one logs a warning naming its replacement.

```
WARN deprecated environment variable: rename it before the next major release
     variable=SQLITE_PATH use=GOMODEL_SQLITE_PATH
```

## Why

The environment namespace is global and flat — shared with every other process in the container. `BASE_PATH`, `HTTP_TIMEOUT`, `STORAGE_TYPE`, and `LOGGING_ENABLED` are land grabs on generic names that say nothing about who owns them, and they collide the first time an operator shares an `env_file` across Compose services or an `envFrom` ConfigMap across containers.

This follows the common convention — prefix what you own, go bare only to join someone else's convention. WordPress (`WORDPRESS_DB_HOST`), Apache (`APACHE_RUN_USER`), Grafana (`GF_SERVER_HTTP_PORT`), Vault (`VAULT_ADDR`), Ollama (`OLLAMA_HOST`). LiteLLM makes the same two-tier split: `LITELLM_MASTER_KEY` for its own config, bare `OPENAI_API_KEY` for the ecosystem.

`GOMODEL_MASTER_KEY` and `GOMODEL_CACHE_DIR` already followed the convention. "Everything bare except two" was the worst state — readers couldn't infer the rule either way.

## What stays bare (permanently)

| | Reason |
|---|---|
| `PORT`, `REDIS_URL` | PaaS platforms inject them. Railway/Heroku set `PORT`; their Redis plugins set `REDIS_URL`, so bare = zero-config wiring. |
| `<PROVIDER>_API_KEY`, `<PROVIDER>_BASE_URL`, `<PROVIDER>_MODELS`, ... | Each vendor's namespace. `OPENAI_API_KEY` is the OpenAI SDK's own name — reading it bare is what makes GoModel drop-in. Generated from provider name + suffix, so it moves as one unit or not at all. |
| `DOCKER_HOST`, `MOCK_*`, `RECORD`, `UPDATE_GOLDEN`, `TEST_*` | Test harness only. |

`POSTGRES_URL` / `MONGODB_URL` are **not** exempt despite looking like ecosystem names — nothing injects them. The real convention is `DATABASE_URL`, which GoModel doesn't read today; accepting it would be a separate feature, not a rename.

## Design notes for review

- **Why `internal/envcompat` and not `config/`:** `HTTP_TIMEOUT` is read twice — by the config struct tags *and* independently by `internal/httpclient`, which can't import `config`.
- **Empty canonical doesn't shadow legacy.** A non-empty value wins, canonical first. An unexpanded `GOMODEL_SQLITE_PATH=` in a compose file alongside a real `SQLITE_PATH` resolves to the real one instead of silently discarding it. Presence is still reported when only empty values are set, so `ok`-bool callers (`RESPONSE_CACHE_SIMPLE_ENABLED=` means "off") keep working.
- **Two resolution paths.** Exact lookup covers the whole tagged config through the single `os.Getenv` in `applyEnvOverridesValue`. A prefix scan covers the four families discovered by walking `os.Environ()`: `SET_BUDGET_*`, `SET_RATE_LIMIT_*`, `SET_PROVIDER_RATE_LIMIT_*`, `TAGGING_HEADER_<N>`.
- **`Scan` sorts by suffix.** Callers resolve suffixes to canonical keys and two suffixes can collide there (`SET_RATE_LIMIT_A_` and `SET_RATE_LIMIT_A` both name path `a`), so iteration order decided the winner — previously dependent on OS ordering.
- **`Scan` reports the legacy spelling**, so companions resolve back through `envcompat`. This is what lets a canonical `GOMODEL_TAGGING_HEADER_1` pair with a legacy `TAGGING_HEADER_1_PREFIX`.
- **`GOMODEL_CONFIG_STRICT`** governs the YAML parse and runs before the tag walker, so it calls `envcompat` at its own site.
- **`LOG_LEVEL` / `LOG_FORMAT` are included** — they live in `run/` rather than `config/` and were missed by the original survey.

## Verification

Beyond unit tests, run against the real binary with mixed spellings:

- `/health` → **200** with bare `PORT=18099` (exempt path intact)
- `GOMODEL_SQLITE_PATH` created the DB at the canonical path (prefix drives real behavior)
- `GOMODEL_LOG_FORMAT=json` took effect
- Only the legacy `LOGGING_LOG_BODIES` warned; canonical vars stayed quiet

`make test-race`, `make lint`, and the hot-path alloc guard all pass.

## Not in this PR

Documentation still shows the legacy spellings. Renaming that surface (`.env.template`, `config.example.yaml`, `README`, `helm/`, compose files) is a large mechanical diff that the exempt rules above make unsafe to do blindly — it's a follow-up. `.env.template`, `CLAUDE.md`, and `docs/dev/2026-07-17_env-prefix-migration.md` document the convention and the full mapping in the meantime.

## Summary by CodeRabbit

* **New Features**
  * Added support for canonical `GOMODEL_` environment-variable names.
  * Legacy unprefixed names still work temporarily and emit one-time migration warnings; canonical values take precedence when both are set.
  * Preserved permanent exceptions for `PORT`, `REDIS_URL`, and provider-scoped variables (with special-case handling where applicable).
  * Environment-based configuration (including caching, tagging headers, MCP servers, virtual models, keyed/rate-limit settings) now consistently honors the canonical naming.
  * Health and readiness probes no longer output logs.
* **Documentation**
  * Added/expanded environment-variable migration guidance, mappings, precedence rules, and exceptions.
* **Tests**
  * Added coverage for canonical/legacy env prefix resolution and override behavior.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,3 +1,17 @@
+# GoModel-defined variables are canonically spelled GOMODEL_<NAME> (e.g.
+# GOMODEL_SQLITE_PATH). The unprefixed spellings shown below still work but are
+# deprecated: each one logs a warning at startup naming its replacement, and
+# they will be removed in a future major release. Setting both spellings
+# resolves to the GOMODEL_ one.
+#
+# Two groups keep their bare names permanently and must not be prefixed:
+#   - PORT and REDIS_URL, which PaaS platforms inject.
+#   - The provider family (OPENAI_API_KEY, ANTHROPIC_API_KEY,
+#     <PROVIDER>_BASE_URL, <PROVIDER>_MODELS, ...), which lives in each
+#     vendor's namespace and is what makes GoModel drop-in compatible.
+#
+# See docs/dev/2026-07-17_env-prefix-migration.md for the full mapping.
+
 # Server Configuration
 # PORT=8080
 # Mount the whole gateway under a path prefix, e.g. https://example.com/g/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,6 +104,17 @@ If this repository is not the official GoModel repository, ask the user whether 
 
 Full reference: `.env.template` and `config/config.yaml`
 
+**Env var naming:** GoModel-defined variables are canonically spelled
+`GOMODEL_<NAME>` (`GOMODEL_SQLITE_PATH`, `GOMODEL_LOGGING_ENABLED`, ...). The
+unprefixed spellings listed below still resolve but are deprecated and warn once
+each at startup; when both are set, the `GOMODEL_` one wins. Two groups keep
+their bare names permanently: `PORT` and `REDIS_URL` (injected by PaaS
+platforms), and the provider family (`OPENAI_API_KEY`, `<PROVIDER>_BASE_URL`,
+`<PROVIDER>_MODELS`, ...), which lives in each vendor's namespace and is what
+makes GoModel drop-in compatible. New variables must take the prefix unless they
+fall in one of those two groups. Details and the full mapping:
+`docs/dev/2026-07-17_env-prefix-migration.md`.
+
 **Key config groups:**
 
 - **Server:**
--- a/config/cache.go
+++ b/config/cache.go
@@ -3,9 +3,10 @@ package config
 import (
 	"fmt"
 	"math"
-	"os"
 	"strconv"
 	"strings"
+
+	"github.com/enterpilot/gomodel/internal/envcompat"
 )
 
 // CacheConfig holds model and response cache configuration.
@@ -262,7 +263,7 @@ func mergeSemanticResponseDefaults(sem *SemanticCacheConfig) {
 }
 
 func applyResponseSimpleEnv(resp *ResponseCacheConfig) error {
-	v, ok := os.LookupEnv("RESPONSE_CACHE_SIMPLE_ENABLED")
+	v, ok := envcompat.Lookup("RESPONSE_CACHE_SIMPLE_ENABLED")
 	if ok && !parseBool(v) {
 		resp.Simple = nil
 		return nil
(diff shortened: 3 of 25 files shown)


SantiagoDePolonia:
Follow-up commit `9b9626d1` closes the findings from the multi-agent review:

- **`Scan` now resolves through the same `lookup` as `Lookup`** — a blank canonical (`GOMODEL_SET_BUDGET_TEAM=` from an unexpanded compose variable) no longer discards a working legacy rule, and the two code paths can no longer diverge. `Entry.Name` is now the spelling that actually supplied the value, so startup errors name a variable that exists in the operator's environment.
- **Whitespace-only canonical values** no longer shadow working legacy values.
- **`OPENROUTER_SITE_URL` / `OPENROUTER_APP_NAME`** (the one missed call site repo-wide) now route through envcompat; canonical spellings resolve, bare ones warn. Documented as judgment calls in the migration doc.
- **`LOG_LEVEL`/`LOG_FORMAT` warnings** are deferred (new `envcompat.Quiet`) until the configured slog handler is installed — a JSON deployment now gets JSON warning lines instead of one unparseable bootstrap-text line. Verified end-to-end.
- **`--health`/`--ready` probes** discard slog output, so periodic healthchecks no longer re-emit the full deprecation warning set every 30s.
- **`GOMODEL_PORT`/`GOMODEL_REDIS_URL`** now warn once that they are not read, instead of silently binding defaults.

New regression tests cover each behavior (blank-canonical Scan fallthrough, whitespace shadowing, Quiet-then-Get warn-once, prefixed-exempt warning, canonical `GOMODEL_OPENROUTER_*` spellings).

PR #566, opened 2026-07-20
feat(providers): configure provider credentials from the dashboard

SantiagoDePolonia:
## Summary

Lets GoModel run and be fully configured **without any provider API-key env vars**. Providers declared via env vars/`config.yaml` keep working exactly as before and stay read-only in the dashboard; a new "Providers" page lets you add, edit, disable, and delete dashboard-managed provider credentials, which apply to the running gateway immediately — no restart.

- **New `provider_credentials` admin store** (SQLite/PostgreSQL/MongoDB), mirroring the existing MCP-servers declarative-shadows-store precedence: a provider name declared via env/config.yaml is always read-only (shown with a "Config" badge), and a dashboard-managed row hot-registers/unregisters into the live registry.
- **`ModelRegistry.UnregisterProvider`** (new — didn't exist before) makes edit/disable/delete work live. `providers.Init` and the registry's refresh path no longer treat "zero providers configured" as a startup failure — that's the whole point of the feature.
- **Admin API**: `GET/PUT /admin/provider-credentials`, `GET /admin/provider-credentials/types`, `DELETE /admin/provider-credentials/:name`. API keys and service-account secrets are redacted as `***` (positional for the key-rotation list); an upsert echoing `***` back preserves the stored value at that position.
- **Dashboard**: new "Providers" page — Type-first form with name auto-suggestion (`{type}`, then `{type}-1`, `{type}-2`, ...), repeatable API-key rows, and less-common fields (Base URL, Models, Vertex/service-account, ...) tucked behind an "Advanced settings" disclosure. Standardized icon-button and toggle-switch shapes (rounded-square instead of circular) and disabled-field styling app-wide along the way.
- **Docs**: `CLAUDE.md`, `docs/providers/overview.mdx`, `.env.template`, `README.md`, regenerated Swagger/OpenAPI.

Three real bugs were found and fixed via actually running the built binary end-to-end (not just unit tests with mocks): `providers.Init` hard-failing on zero providers, the registry's `Refresh` erroring on zero remaining providers, and `Upsert`/`Delete` conflating a provider's own network/credential failure with a save failure (a wrong API key would return a scary error instead of just showing "Unhealthy").

## Test plan

- [x] `go build ./...` and `go vet ./...`
- [x] `go test ./internal/...` (full suite, including new registry/credentials/admin tests)
- [x] `make test-dashboard` (499 JS unit tests)
- [x] `make lint` / pre-commit hooks all green
- [x] Live end-to-end smoke test: built `cmd/gomodel`, ran with zero provider env vars (boots successfully), exercised create/update/delete against `/admin/provider-credentials`, verified `"***"` redaction round-trips correctly, verified env-only declared providers (e.g. `OLLAMA_BASE_URL` with no config.yaml entry) are correctly protected as read-only
- [x] Verified dashboard page renders at `/admin/dashboard/providers-config`

## Summary by CodeRabbit

* **New Features**
  * Added an admin dashboard page and REST API to manage provider credentials (create/edit, enable/disable, delete) with immediate routing updates.
  * Providers can now be configured after startup, and the gateway can boot with no provider API keys set.
  * Added Anthropic Messages batch endpoints (including JSONL results).
  * Updated MCP documentation and naming to use slug-based conventions.
* **Documentation**
  * Updated guidance on provider credential precedence, managed/read-only behavior, and secret redaction.
  * Documented `LOGGING_RETENTION_DAYS` and `stop_sequence`.
* **Bug Fixes**
  * Improved startup behavior when no providers are configured.


SantiagoDePolonia:
Reviewed all bot comments critically (not just autofixed). Verified each CodeRabbit finding against the actual code before acting:

**Fixed** (all in caa1abb9):
- **Major** — `CredentialsService.Upsert` unregistered the live provider *before* validating the replacement. An edit that failed to resolve (e.g. an accidental empty API key on an already-working provider) took it fully offline with no way back except another edit. Confirmed by tracing the code — real bug. Fixed by resolving/constructing the replacement first (`buildProvider`) and only swapping it into the registry (`install`) on success; added a regression test (`TestCredentialsService_UpsertKeepsThePreviousProviderLiveWhenTheEditIsUnresolvable`) that would have caught this.
- **Minor** — no validation rejected `/` in provider names, which would break `"name/model"` selector routing. Added the check at both the service and handler layer (handler gets a clean 400 instead of a 502-wrapped service error), with tests at both layers.
- **Minor** — `UpsertProviderCredential`/`DeleteProviderCredential` bypassed the admin handler's `mutationMu`, unlike guardrails/tagging/workflows. Added it — closes a real (if narrow) window where two concurrent edits of the same provider name could interleave between the redaction-merge read and the store write.
- **Minor** — the credential store wasn't closed if `NewCredentialsService` failed after `createCredentialStore` succeeded. Fixed; note this mirrors a pre-existing (unreported) gap in `mcpgateway`'s equivalent wiring — out of scope here, left as-is.

**Deliberately not changed:**
- The "Docstring Coverage" pre-merge check (31.94% vs. an 80% threshold) — this repo's actual style guidance (`CLAUDE.md`) is the opposite: no comments unless the *why* is non-obvious. Chasing that metric would work against the project's own conventions, so I left it.
- Codecov's 0% coverage on `credentials_store_postgresql.go`/`credentials_store_mongodb.go`/`credentials_wire.go` — matches the existing, already-accepted pattern for `mcpgateway`'s Postgres/Mongo stores (only SQLite gets a unit test there either); app-wiring plumbing in `app.go`/`credentials_wire.go` was instead verified by actually building and running the binary end-to-end with zero provider env vars.

PR #582, opened 2026-07-24
feat(dashboard): unify mutation feedback into global flash toasts

SantiagoDePolonia:
## What

Replaces the dashboard's per-page notice/error banner pattern (hand-rolled $state fields, inline alert markup, and page-specific gating repeated across 19 files) with a single flash-message mechanism:

- New `flash` store (`src/lib/stores/flash.svelte.js`): `flash.success(msg)` / `flash.error(msg)` queue toasts with timed auto-dismiss (5s success / 8s error), manual dismiss, and dedup (an identical toast still on screen restarts instead of stacking).
- New `FlashMessages` organism mounted once in `App.svelte`: fixed bottom-right stack, rendered above modal shells so feedback fired from a dialog stays visible, with consistent `role="status"`/`role="alert"` + `aria-live` in one place.
- All 10 page stores (budgets, rate limits, MCP servers, provider credentials, guardrails, workflows, virtual models, pricing overrides, failover, API keys) and 4 settings sections (budget settings, tagging, pricing recalculation, runtime refresh) migrated; every per-store `notice` field is deleted.
- Toasts fire immediately when the mutation response arrives — before the follow-up list refetch — so feedback is instant instead of trailing the refetch by seconds.

## User-visible impact

- Mutation feedback (saves, deletes, resets, toggles, reconnects) now appears as bottom-right toasts that auto-dismiss, instead of inline banners that persisted until the next action and were sometimes hidden by gating conditions (e.g. the failover editor's "Generated N fallback models" message previously rendered nowhere visible on the Models page).
- Deliberately unchanged: load failures, in-form validation/submit errors, "feature is unavailable" capability warnings, confirm-dialog errors, and the one-time issued-API-key banner remain inline — transient toasts would be wrong for persistent state.
- "Generated N drafts" no longer duplicates the failover drafts modal's own list/empty state.

## Testing

- svelte-check: 0 errors / 0 warnings; `make test-dashboard`: 316/316.
- Live headless-browser (CDP) verification: toast text/role/styling, 5s auto-dismiss, manual dismiss, and instant appearance on budget create; embedded dist rebuilt and in sync.

## Summary by CodeRabbit

* **New Features**
  * Added global toast notifications for dashboard success and error messages, including fixed toast stacking, auto-dismiss, animations, accessibility live regions, and manual dismissal.
* **Bug Fixes**
  * Unified user feedback across settings, budgets, guardrails, models, providers, rate limits, workflows, failover, and API key actions by replacing per-page success “notice” banners with flash toasts.
  * Improved consistency: identical messages are de-duplicated, and success banners are removed where they were previously incorrectly shown/hidden during form interactions.


SantiagoDePolonia:
Addressed both review findings in b1431c88:

- **Refetch failures misreport mutations** — post-success list refreshes are now fire-and-forget (`void`) rather than awaited inside the mutation's try block (26 sites), so a refresh failure can only surface as the list's own load state, never as mutation feedback. Toast timing is unaffected since feedback already fired before the refresh.
- **Successful saves retain load errors** — `BudgetSettings` and `TaggingSettings` now clear a retained load error on a successful save, since the save proves the endpoint works and returned fresh data.

PR #584, opened 2026-07-24
feat(admin): restrict dashboard and admin API to selected API keys

### LINKED ISSUE #583: Feature: API key restriction - only the admin key or selected keys should be able to access the dashboard
As titled.

SantiagoDePolonia:
## Summary

Managed API keys now carry a `dashboard_access` flag (default: **false**). Only the master key and keys with the flag can call `/admin/*` — other managed keys get `403` with code `dashboard_access_denied`. Model endpoints and the self-service `GET /v1/usage` stay open to every key.

Closes #583

## What changed

- **authkeys**: `DashboardAccess` on `AuthKey`/`CreateInput`/`AuthenticationResult`; column migrations for SQLite/PostgreSQL/MongoDB; `UpdateDashboardAccess` store + service methods.
- **admin API**: `dashboard_access` accepted on `POST /admin/auth-keys`; new `PUT /admin/auth-keys/{id}/dashboard-access` to grant/revoke on existing keys.
- **server**: `AdminAccessMiddleware` gates the `/admin` group and the legacy `/admin/api/v1` alias. The auth middleware records the managed key's access in the request context; master-key requests and the no-master-key lockout-recovery path (auth skipped on `/admin/*`) are untouched.
- **dashboard**: create-form checkbox, a Dashboard Access table column with helper text (denied keys lose `/admin` API access, `/v1/usage` stays available), a grant/revoke row action, and the auth dialog reopens with a specific message when a key without access is used.
- **docs**: `docs/advanced/admin-endpoints.mdx` authentication section + CLAUDE.md config reference.

## Breaking change

Existing managed keys are **denied by default** (deliberate, per the issue): operators who currently use a managed key in the dashboard must grant it dashboard access with the master key (`PUT /admin/auth-keys/{id}/dashboard-access`) or use the master key.

## Tests

- Full-server gate tests: 403 + `dashboard_access_denied` for unflagged keys on both `/admin` and the legacy alias; flagged key and master key pass; `/v1/usage` and `/v1/models` unaffected; lockout-recovery path stays open.
- authkeys service/store round-trips (including the SQLite migration on an existing DB), admin handler tests, and a JS test for the create payload.
- `go test ./...` green; dashboard unit tests green; dist rebuilt and in sync.

## Summary by CodeRabbit

* **New Features**
  * Added per-auth-key “Dashboard access” control, including a new admin endpoint to grant or revoke it.
  * Admin UI now supports toggling dashboard access from the auth key list and setting it during key creation.

* **Documentation**
  * Expanded admin endpoint authentication guidance, including the `dashboard_access_denied` 403 behavior and lockout-recovery boundaries.

* **Bug Fixes**
  * Improved admin authentication error messaging so denied dashboard access displays clearer, specific UI text.

* **Tests**
  * Added coverage for dashboard-access persistence and admin access gating behavior.


SantiagoDePolonia:
Addressed all five review findings in e04df28b:

- **Omitted/null `dashboard_access` on PUT**: request field is now `*bool`; `{}` and `{"dashboard_access": null}` return 400 instead of silently revoking (tests added).
- **Toggle disabled state**: all dashboard-access toggles are disabled while one update is in flight, matching the store's single-flight guard.
- **Legacy alias positive path**: replaced the two single-path tests with a table-driven allowed-credential matrix covering `/admin/api/v1/*` for both the master key and a flagged key.
- **SQLite test helper**: `assertAccess` now asserts `List` returns exactly the expected key set (count + unexpected-ID check).
- **CLAUDE.md claim**: scoped to admin REST endpoints, with the public dashboard shell/static assets called out explicitly.

PR #585, opened 2026-07-25
refactor(dashboard): dead CSS, duplication and oversized components

SantiagoDePolonia:
Housekeeping pass over the Svelte dashboard: **69 files, +689 / −2657** in source (plus the regenerated `dist/`). No intended behavior change apart from one fix, called out below.

## Dead CSS

`dashboard.css` drops 2618 → 2400 lines. Removed 14 unreachable rule groups: an Alpine.js `[x-cloak]` leftover, an unwired tab-transition, five `budget-bar-fill-period-*` colours, two `.is-editor-open` compounds nothing sets, an empty `@media` block, `.editor-modal-shell-wide` (two components carry comments saying the `Modal` atom has no wide variant — it doesn't), and more.

Svelte's unused-selector warning can't see any of these: they're either global or sit in components whose `class` attribute is built from an expression, which makes the compiler conservative.

## One fix

`AuditPane`'s request/response direction arrows are meant to be tinted accent/blue (there's a comment saying so), but the global modifier `.audit-pane-icon-request` (0,1,0) lost to the scoped base `.audit-pane-icon.svelte-hash` (0,2,0) and **both rendered muted grey**. Confirmed against the previously committed compiled CSS. The modifiers now live in the owning component as compound `:global` selectors, so they stay reachable for dynamically composed classes *and* outrank the base. This is the one visible change in the PR. The cascade trap is now documented in `CONVENTIONS.md`.

## Extracted

| New | Replaces |
|---|---|
| `NoDataIllustration` | a 60-line SVG copy-pasted byte-identically into 3 pages |
| `FilterInput` | the search-icon + input block repeated in 11 pages |
| `CopyButton` | the copy/check icon-swap button in 3 places |
| `AuditEntry{Summary,Metadata}` + `AuditPaneTabs` | `AuditEntryRow`, 622 → 71 lines |
| `WorkflowIdBadge` | the id pill inside `WorkflowChart` |
| `DatePickerCalendar` + `datePickerLogic` | `DatePicker`, 576 → 293 |

Deleted `pages/models/TableIcon.svelte` — a hand-rolled name→SVG switch duplicating the lucide `Icon` atom. The rest of the app already rendered the same table actions via `Icon`, so `models/` had visibly different pencil/trash glyphs from API Keys and MCP Servers.

## Deduplicated

A scan for identical function bodies found 11 groups; all are now zero. Notably four byte-identical `{error:{message}}` extractors (plus four near-identical test blocks), the chart tooltip/tick fragments in two modules, and the 11-colour palette in three places with `labelColor`'s djb2 hash duplicated verbatim.

New `$lib/api/errors.js` exists rather than living in `client.js` because `client.js` imports rune-based stores, which pure page logic and `node:test` can't load.

## Renamed `.pagination-btn` → `.btn`

It was the app's base button class on ~37 elements unrelated to pagination — every primary action read `class="pagination-btn pagination-btn-primary pagination-btn-with-icon"`. Renamed across 36 files (157 occurrences); `.pagination` (the container) is unchanged. Token-exact, and `.btn-danger` doesn't collide with the pre-existing `.table-action-btn-danger`.

## Verification

- `svelte-check`: 0 errors, 0 warnings
- tests: 317 → **330** (new suites for the extracted date math, debounce and error helpers)
- `npm run build` clean; embedded `dist/` regenerated and in sync
- Rendered against a live gateway: Overview, Models, Audit Logs (including an expanded entry — workflow chart, request/response tabs, metadata strip), Usage, Settings, API Keys

The new date-picker tests immediately caught a bug I'd introduced while extracting `calendarDays`: `for (let d = 1; d <= 42 - days.length; d++)` re-evaluates its bound each iteration, so the grid came out at 38–40 cells instead of 42. Fixed.

## Review notes

Reviewing commit-by-commit is easier than the combined diff — the second commit is a pure mechanical rename.

Three files remain slightly over the 400-line guideline (`WorkflowChart` 575, `Sidebar` 435, `ProviderStatusCard` 429). The remainder in each is a flat, commented status/state CSS matrix; splitting is blocked because the state classes are computed strings, so a child component's scoped rules would be pruned — now noted in `CONVENTIONS.md`. The light/dark token duplication also has to stay: `light-dark()` would break `chartTheme.cssVar()`, which reads those variables back with `getComputedStyle` — now explained at the top of `dashboard.css`, replacing a TODO that blamed the missing CSS preprocessor.

## Summary by CodeRabbit

* **New Features**
  * Added reusable `FilterInput`, `NoDataIllustration`, `CopyButton`, and `ThemeToggle` components (plus a workflow ID badge).
  * Expanded the audit logs UI with dedicated metadata/summary and request/response tabs.
  * Rebuilt the date picker with a calendar sub-component and shared UTC calendar math.
* **Bug Fixes**
  * Improved API/admin error-message extraction via shared helpers (with added test coverage).
* **Style**
  * Standardized buttons/icons, updated sidebar/modal styling, and improved shared chart theming and CSS scoping guidance.
* **Tests**
  * Added/updated unit tests for the new shared utilities and removed obsolete helper tests.


SantiagoDePolonia:
Thanks — went through all five. Four accepted, one accepted in part, in 67806a7.

**`errorPayloadMessage` whitespace** — accepted. Confirmed: `"   "` was returned verbatim while its sibling `errorMessage` trims. They sit side by side in `errors.js` under a comment saying they do the same thing, so the difference was a trap. Now trims, with a test. (For the record this matches the four page-local helpers it replaced — none trimmed — so it's not a regression from this PR, just newly visible once they were consolidated.)

**`NoDataIllustration` with `label=""`** — accepted. Now `role`/`aria-label` are dropped and `aria-hidden` set. Worth noting the option is currently unused (all three call sites render the labelled form), so this is the documented behaviour catching up with the code rather than a live bug.

**`aria-pressed` on the theme buttons** — accepted, and I matched the house pattern: `SegmentedControl.svelte` already does `role="group"` + `aria-pressed` for exclusive option groups, so the theme pill now does the same. Selection state was previously conveyed by background colour alone.

**Mobile theme button label** — accepted. It cycles rather than selects, so announcing `"Light theme"` described a state, not a control. Now `"Change theme (currently Light theme)"`.

**Collapse `import` + `export {}` into `export ... from`** — accepted for 2 of 7 modules, declined for the rest. `export { x } from` creates no *local* binding, so any module that also calls the symbol internally breaks at runtime — that was the original form here and the test suite caught it. Five modules call theirs internally (`mcp-servers`, `providersConfigLogic`, `failover-logic`, `mcpOverviewLogic`, `pricing-logic`); `authKeysLogic` and `usage-helpers` don't, so those two are collapsed.

**`CONVENTIONS.md` utils** — accepted in part. I widened the `format.js` one-liner to cover date params and comma lists rather than re-listing every export: the previous commit deliberately replaced that per-export catalogue with an index, because it drifted and nothing enforced it. But the underlying point stands — two pages had independently reimplemented `splitCommaList`, so the index does need to hint that it lives there.

`svelte-check` 0/0, 331 tests green, `dist/` in sync.

PR #586, opened 2026-07-25
refactor(storage): serve both SQL backends from one store implementation

SantiagoDePolonia:
Collapses the triplicated SQL store layer onto one implementation per domain, and fixes the connection and shutdown wiring that duplication was hiding.

Full write-up, including what is deliberately left and why: `docs/dev/2026-07-25_backend-refactor-survey.md`.

## Why

Every persisted domain shipped three near-identical stores plus a factory — **18,259 lines, 15.5% of the hand-written backend**. Normalising placeholders and type names, the SQLite and PostgreSQL halves were 50–83% textually identical; the entire difference was six mechanical things (driver handle, `?` vs `$n`, no-rows sentinel, `RowsAffected` signature, `Exec` arity, DDL type names).

The coverage asymmetry mattered more than the line count:

| Backend | store test files | implementations | untested |
|---|---:|---:|---:|
| SQLite | 22 | 22 | 0 |
| PostgreSQL | **4** | 22 | **18** |
| MongoDB | 9 | 22 | 13 |

…and those four PostgreSQL "tests" only asserted generated SQL strings. **No PostgreSQL store code executed against a database anywhere in the suite.** It had drifted accordingly: `failover` normalised padded primary keys through two independently written migrations, only one of them ever verified.

## What changed

`internal/storage/sqlx` absorbs exactly those six differences. **16 of 17 domains** now have one store implementation.

| | before | after |
|---|---:|---:|
| SQLite store code | 5,570 | 1,705 |
| PostgreSQL store code | 5,015 | 1,448 |
| Factories | 2,232 | 1,762 |
| Unified `store_sql*.go` | — | 3,948 |
| `internal/storage/sqlx` | — | 820 |

**PostgreSQL subtests executing against a real database: 0 → 91**, wired into CI with a `postgres:18-alpine` service.

Two things the adapter deliberately does **not** abstract, both settled by probing the drivers rather than assuming:

- **Value binding.** Both drivers already agree — a Go `bool` binds to SQLite `INTEGER` and PG `BOOLEAN`, an `INTEGER` scans into `*bool`, `TEXT`/`JSON`/`JSONB` all scan into `[]byte`. The `boolToSQLite` helpers and the `sql.NullString`-vs-`*string` split were accidental divergence, never dialect requirements.
- **Genuinely dialect-specific SQL**, kept behind a `Dialect()` check and labelled: `conversationstore`'s JSON mutations (JSON1 functions vs jsonb operators — must be one server-side `UPDATE` or concurrent Responses turns clobber each other), the `failover`/`ratelimit` legacy migrations, `budget.SumUsageCost`, and `auditlog`'s timestamp binding.

**Also fixed:** the app opened **up to 15 connections to the same database** when audit logging and usage tracking were both disabled — each subsystem had a `New` that called `storage.New` itself. Now one handle opened in `app.New`; 14 duplicate constructors and 19 `Result.Storage` fields deleted. `Shutdown`'s 15 copies of the same close block become an ordered list with the connection closed last: 194 → 102 lines.

## Behaviour changes worth reviewing

1. **`workflow_versions.created_at` on PostgreSQL** was `TIMESTAMPTZ` while SQLite used INTEGER unix seconds — the only table where the backends disagreed on representation. It now converges on unix seconds (what every other table already used on both engines) via an in-place `EXTRACT(EPOCH FROM ...)` migration in `NewSQLStore`. **This is the one change that touches existing data.**
2. **One connection instead of N.** A fix, but observable.
3. **`InTx` on SQLite now uses `BEGIN IMMEDIATE`**, taking the write lock up front instead of lazily.

## Verification

Every commit passed the full pre-commit gate (`make test-race`, `make lint`, `make fix-check`) across all build tags. Beyond that, the whole suite ran against a live PostgreSQL 18, and the built binary was smoke-tested on both backends — boot, `/health`, an auth-gated `/v1/models`, graceful shutdown.

That smoke test is what caught the `workflow_versions` break, and it points at the useful lesson: **a conformance suite over fresh schemas proves the dialects agree but says nothing about upgrading a database written by an older release.** Migration paths now have tests that start from the legacy table shape (`guardrails`, `mcpgateway`, `failover`, `ratelimit`, `workflows`, `auditlog`).

## Not done, deliberately

The **`usage` store** and both **analytics readers**. The readers are the one place where the divergence is real analytics SQL, not duplication. The `usage` store is coupled to its reader — `RecalculatePricing` builds rows from the reader's condition builders, and the two implementations differ *semantically* (SQLite paginates by `id`; PostgreSQL takes every matching row with `FOR UPDATE`), so unifying now would change behaviour on one engine. It should follow the reader rather than lead it.

Also deferred with reasons in §7: the `CacheModeCached` four-owner cleanup, the config-shadows-store precedence question (9 subsystems, 3 different semantics — a product decision), and MongoDB, which stays hand-written by decision.

## Summary by CodeRabbit

* **New Features**
  * Added unified SQL-backed persistence across core data stores, including automatic schema migrations and legacy-shape compatibility.
* **Bug Fixes**
  * Improved startup/shutdown reliability with safer cleanup of optional components and verified idempotent shutdown.
  * Refreshed storage wiring so usage/admin dashboard behavior consistently uses the shared storage connection.
* **Tests**
  * Expanded CI to run store tests against both SQLite and PostgreSQL.
  * Added broader SQL conformance and store integration test coverage, plus an upgrade-compatibility E2E script.
* **Documentation**
  * Documented the one-way PostgreSQL upgrade change for `workflow_versions.created_at` (unix seconds, UTC, floored).


SantiagoDePolonia:
Thanks — eight of these were right and are fixed in dbf52af4. Three I'm declining, with reasoning:

**`internal/failover/store_sql_migrate.go` — "SQLite skips the trim-only case" (Major).** The asymmetry is real but it isn't a gap, because the two engines had different histories. The SQLite migration has *always* trimmed: it rebuilds the table and copies rows through `SELECT TRIM(primary_model) ... WHERE TRIM(COALESCE(...)) <> ''`, so there is no "earlier non-trimming version" for SQLite to heal from. PostgreSQL renames the column in place, so an earlier release could and did leave padded keys behind — which is exactly why its migration trims independently of the rename. `Upsert` trims on the way in on both engines, so padded keys cannot appear afterwards either.

Adding an unconditional trim on SQLite would also not be free: if two rows differ only by padding it collides on the primary key at startup. I kept both paths verbatim on purpose — they run against databases in the field where a subtle change would silently orphan rules.

**`internal/storage/sqlx/dialect.go` — "Keep `TypeJSONText` → `TEXT` on PostgreSQL".** This would break the thing it's trying to protect. `main`'s PostgreSQL store already declared `fallback_models JSONB NOT NULL DEFAULT '[]'::jsonb` (`internal/failover/store_postgresql.go:28`), and the SQLite one declared `TEXT`. `TypeJSONText` → `JSONB`/`TEXT` reproduces both exactly; mapping it to `TEXT` on PostgreSQL is what would change fresh installs relative to today. Two JSON tokens exist precisely because some SQLite tables declared `JSON` and others `TEXT`, and both had to be preserved.

The fresh-vs-migrated column type difference you describe predates this PR (same `CREATE TABLE` plus the same rename), and is benign either way: `encodeTargets` binds a JSON string, which both column types accept, and `decodeTargets` takes `[]byte`, which both return.

**`.github/workflows/test.yml` — pin the image to a digest.** Declining for an ephemeral CI-only Postgres that holds throwaway schemas: a digest pin freezes it against upstream security updates and needs manual bumps, which costs more than it buys here.

SantiagoDePolonia:
Reviewed the five findings on the last two commits. **Four were right and are fixed in a06026b1; one is declined.**

**Fixed**

- **`EXTRACT(EPOCH FROM created_at)::bigint` rounds.** This was the substantive one. Verified against a real database: `1700000000.600` migrated to `1700000001`, one second into the future — and disagreeing with the truncation `time.Unix` performs on every row the store writes afterwards, so a migrated row and a freshly written one behaved differently. Now `FLOOR(EXTRACT(...))::bigint`. `TestNewSQLStoreConvertsTimestamptzCreatedAt` seeds `.4` and `.6` rows either side of the boundary; it fails on the previous cast and passes on the fix.
- **`upgrade-compat.sh` only rebuilt `bin/gomodel` when missing.** Correct, and it is exactly the trap this repo has hit before — a stale binary reports a pass for code that predates the change under test. It now always rebuilds.
- **The harness never write-checked filestore**, while the survey doc claims every domain accepts writes post-upgrade. Added the upload; 99 checks green across all three backends.
- **Docs wording.** "No action needed" was misleading next to a one-way migration, and "keeps its instant" was wrong given the rounding. Now: no manual migration required, take a backup if you may need to roll back, and values are floored to whole seconds.

**Declined**

- **Sanitizing `\ . " $` and slicing on rune boundaries in `mongoTestDatabaseName`.** The input is `t.Name()`, which is a Go function identifier plus the subtest names this file passes to `t.Run` (`memory`, `sql`, `mongo`, and the dialect names). None of those characters can appear, and the string is ASCII, so the byte slice cannot split a rune. Adding the table-driven coverage would only test inputs the function cannot receive.

PR #595, opened 2026-07-26
fix(auditlog): keep request revisions on streamed entries

SantiagoDePolonia:
A streamed request is persisted from the `CreateStreamEntry` copy — the base entry is never written — so an ingress rewrite chain recorded by `EnrichEntryWithRequestRevision` was lost the moment the request streamed. `CreateStreamEntry` builds `LogData` with a field whitelist, and `RequestRevisions` was not on it.

The visible effect: the "Rewritten" audit pane was empty for every successful streamed request, while the same request on the non-streamed error path kept its snapshot.

## Evidence

Successful chat completions in a local database, cross-referenced against whether a request rewriter actually fired:

| stream | revision | rewriter fired | n |
|---|---|---|---|
| 0 | no | no | 5 |
| 0 | yes | no | 4 |
| 0 | yes | **yes** | 2 |
| 1 | **no** | **yes** | **157** |
| 1 | no | no | 73 |

The bug row — *fired but no revision* — exists only for streamed requests. Non-streaming never lost one, because `CreateStreamEntry` has only two callers and both are streaming paths.

## Changes

- `CreateStreamEntry` copies `RequestRevisions` (through `slices.Clone`, so the copy owns its slice) and `RequestBodyTooBigToHandle`.
- A reflection-based drift guard walks `LogData` and fails when any request-side field does not survive the whitelist copy. Response-side fields are exempted by name — that list is the only thing a future field has to be added to, instead of silently vanishing the way `RequestRevisions` did.

## Operational note

Streamed requests will now each persist a revision, and a revision carries a full copy of the rewritten body (gated by `LogBodies`, capped at `MaxBodyCapture` = 1 MiB). Audit storage will grow faster than before: from storing ~1% of streamed rewrites to 100% of them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming audit log entries now correctly preserve request revision details, including rewrite information, byte counts, and saved tokens.
  * Additional request metadata—such as oversized-body indicators and ingress rewrite chains—is retained accurately, without sharing backing storage where applicable.
* **Tests**
  * Added regression tests to ensure request-side log fields are copied reliably (including nil vs empty behavior) and that no request-side fields are unintentionally dropped.

### DIFF
--- a/internal/auditlog/stream_entry_request_fields_test.go
+++ b/internal/auditlog/stream_entry_request_fields_test.go
@@ -58,11 +58,29 @@ func TestCreateStreamEntryPreservesRequestRevisions(t *testing.T) {
 		t.Error("RequestBodyTooBigToHandle dropped")
 	}
 
-	// The copy must own its slice, so later appends to the base entry cannot
-	// reach into the entry the observer is writing.
-	base.Data.RequestRevisions = append(base.Data.RequestRevisions, RequestRevisionSnapshot{Seq: 2})
-	if len(streamEntry.Data.RequestRevisions) != 1 {
+	// The copy must own its slice. Appending to the base entry would not show
+	// that: the source literal has no spare capacity, so append reallocates and
+	// leaves the copy alone whether or not the two share an array. Writing
+	// through an existing element is what actually distinguishes them.
+	base.Data.RequestRevisions[0].Rewriter = "mutated"
+	if streamEntry.Data.RequestRevisions[0].Rewriter != "pro-token-compression" {
 		t.Error("stream entry shares its revision backing array with the base entry")
+	}
+}
+
+// The nil case is the one the stream observer sees most often — most requests
+// carry no rewriter — and it must stay nil rather than becoming an empty slice,
+// so a streamed entry without rewrites serializes the same as it always did.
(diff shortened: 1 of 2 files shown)


SantiagoDePolonia:
Thanks — the aliasing point was correct and it was a real hole in the test, not just a style nit.

I confirmed it before fixing: swapping `copyRequestRevisions` for a plain shallow assignment left `TestCreateStreamEntryPreservesRequestRevisions` **passing**. The source literal has no spare capacity, so `append` reallocates regardless of sharing. Now it writes through an existing element, and that same shallow assignment fails it:

```
--- FAIL: TestCreateStreamEntryPreservesRequestRevisions
    stream entry shares its revision backing array with the base entry
```

I also added the absent-revisions case, which is the common path — most streamed requests carry no rewriter, and the field must stay `nil` rather than becoming an empty slice.

On the table-driven suggestion: skipping it. The two cases here assert different things (one checks field-by-field fidelity plus slice ownership, the other checks a nil stays nil), so folding them into a table would mean a case struct of mostly-unused fields and a body of conditionals — less legible than two named tests, not more. The reflection drift guard already provides the exhaustive per-field coverage a table would be reaching for.

PR #596, opened 2026-07-26
fix(lifecycle): finish shutdown before the process exits

SantiagoDePolonia:
Ctrl+C on `make run` ended in an unexplained `failed to shut down server within given timeout` followed by `make: *** [run] Error 1`. Three independent causes, and the one that mattered most was not the one in the error message.

## 1. The teardown never finished

`Run` returned as soon as `Start` returned, while the goroutine doing the teardown was still working. The process exits the moment `Run` returns, so every closer the teardown had not reached yet — buffered usage records, buffered audit records, the database handle — was dropped. `application shutdown complete` was never reached on any interrupt, and this happened on **every** Ctrl+C, not only the ones that logged an error.

`Run` now waits for that goroutine. The teardown also runs when the server stops without a signal, so a server that dies on its own still releases its resources; `App.Shutdown` is idempotent, which keeps that harmless on the failed-start path.

## 2. The scary error was Echo's, not ours

Nothing configured `GracefulTimeout`, so Echo used its implicit 10s default and reported the cutoff through its own logger as a bare `context deadline exceeded`. It fires whenever any request is in flight — one open stream is enough.

The window is now set explicitly and documented against the shutdown budget that has to contain it, and the cutoff is reported as the routine event it is. The value stays at 10s deliberately: no drain window covers a streamed model response, so lengthening it would only make Ctrl+C slower without changing the outcome.

## 3. `make: Error 1` is `go run`

`go run` exits 1 when interrupted no matter how cleanly the child exits. Verified against a twelve-line program that catches SIGINT and returns 0:

```
[direct binary] parent_exit=0 output=ready|clean exit 0|
[go run]        parent_exit=1 output=ready|clean exit 0|
```

`make run` now builds and `exec`s, which also puts the gateway directly under make's signal handling instead of behind a supervisor.

## Verification

Reproduced with a request deliberately held in flight, before and after:

```
before:  shutting down application... → ERROR failed to shut down server within given timeout → exit (no "complete")
after:   shutting down application... → WARN closing requests still in flight at the shutdown deadline → application shutdown complete
```

`make run` + SIGINT now exits 0. Both new lifecycle tests fail against the old code (`serveUntilShutdown returned mid-teardown`, `shutdownCalls = 0, want 1`) and pass with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown to allow in-flight requests to finish before cleanup proceeds.
  * Ensured shutdown and data flushing complete before the application exits.
  * Added warnings when requests remain active beyond the shutdown deadline.
  * Improved handling when the server stops unexpectedly or encounters a startup failure.

* **Chores**
  * Updated the local run workflow to build and execute the application consistently, with improved signal handling for Ctrl+C.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ build:
 # handling instead of behind a supervisor.
 run:
 	go build -tags=swagger -ldflags '$(LDFLAGS)' -o bin/gomodel ./cmd/gomodel
-	LOG_LEVEL=$(LOG_LEVEL) SWAGGER_ENABLED=$(SWAGGER_ENABLED) exec ./bin/gomodel
+	LOG_LEVEL="$(LOG_LEVEL)" SWAGGER_ENABLED="$(SWAGGER_ENABLED)" exec ./bin/gomodel
 
 # Seed the local SQLite database and start GoModel with a populated dashboard.
 demo: seed-demo-data
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,16 @@ install-tools:
 
 build:
 	go build -ldflags '$(LDFLAGS)' -o bin/gomodel ./cmd/gomodel
-# Run the application
+# Run the application.
+#
+# Built and exec'd rather than `go run`: `go run` exits 1 when it is
+# interrupted, even after the program it supervises shuts down cleanly, so
+# Ctrl+C always ended in a bogus "make: *** [run] Error 1". exec replaces the
+# recipe shell, which also puts the gateway directly under make's signal
+# handling instead of behind a supervisor.
 run:
-	LOG_LEVEL=$(LOG_LEVEL) SWAGGER_ENABLED=$(SWAGGER_ENABLED) go run -tags=swagger ./cmd/gomodel
+	go build -tags=swagger -ldflags '$(LDFLAGS)' -o bin/gomodel ./cmd/gomodel
+	LOG_LEVEL=$(LOG_LEVEL) SWAGGER_ENABLED=$(SWAGGER_ENABLED) exec ./bin/gomodel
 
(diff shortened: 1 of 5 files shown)


SantiagoDePolonia:
All four addressed.

**Double shutdown on the failed-Start path** — agreed, and taken further than "track whether teardown already ran": `startApplication` is deleted and `serveUntilShutdown` is now the sole caller of `shutdownApplication`. Every exit converges on one teardown with one `shutdownTimeout` budget, which drops the dependency on `App.Shutdown` idempotency instead of relying on it. `TestServeUntilShutdown_TearsDownOnceAfterAFailedStart` asserts `shutdownCalls == 1`.

One behavior change worth calling out: a teardown that also fails is now logged rather than joined onto the returned error. The start error is what the operator needs to see and what sets the exit code, and the shutdown error still reaches the log.

**Cross-package invariant** — fair; the comment was load-bearing and unchecked. `GracefulDrainTimeout` is exported and `TestGracefulDrainFitsInsideTheShutdownBudget` asserts both that it is shorter than `shutdownTimeout` and that enough budget remains for the flushes that follow the drain. Confirmed it fails in both directions:

```
30s: GracefulDrainTimeout = 30s must be shorter than shutdownTimeout = 30s
26s: only 4s left for flushing after the drain
10s: ok
```

**Makefile quoting** — applied (and to the pro repo's equivalent recipe).

**Shutdown-count assertion on the start-failure test** — folded into the first item.

Re-verified end to end after the refactor, with a request deliberately held in flight:

```
shutting down application...
WARN closing requests still in flight at the shutdown deadline graceful_timeout=10s
application shutdown complete
exit 0
```

PR #597, opened 2026-07-26
test(e2e): cover label-scoped budgets in the release curl matrix

SantiagoDePolonia:
Adds S197–S204 to the release curl matrix (196 → 204), covering budgets scoped to a request label rather than a `user_path` subtree, as shipped in #590.

| ID | Covers |
|---|---|
| S197 | Admin CRUD round-trip plus validation negatives: `user_path` alongside `subject`, missing subject, unknown scope, delete-absent 404 |
| S198 | Tagging-header enforcement: first request charged, second blocked with `budget_exceeded`, and a different-case spelling **not** blocked — label matching never folds case |
| S199 | One request carrying two labels charges both budgets |
| S200 | A managed key's labels charge a label budget with no tagging rule configured |
| S201 | A request matched by both a `user_path` and a `label` budget |
| S202 | `reset-one` clears a label budget's spend |
| S203 | PostgreSQL and MongoDB parity |
| S204 | Auth gating on the admin budgets API for label scope |

## Two assertions brought up to their prose

While verifying these against the implementation I found two scenarios whose descriptions promised more than the code checked. Both are now asserted:

- **S199** said the block names "the exhausted label rather than the healthy one", but only checked the exhausted one was present. It now also checks the healthy label is absent.
- **S201** said the user-path budget "still has ample room", but never looked at it. It now checks that budget was charged by the same request *and* still has room — which is what makes it evidence that the label budget alone caused the block.

## Why S200 runs on the auth-enabled gateway

On a gateway with no `GOMODEL_MASTER_KEY`, creating a managed key switches every endpoint — `/v1/*` included — to require bearer auth from then on, and managed keys have no delete endpoint (`deactivate` does not undo the switch, which counts stored keys rather than active ones). Registering one on the open SQLite gateway would leave it locked down for every later scenario. The auth-enabled gateway also has the exact response cache on, so both chat replies carry `$QA_BUDGET_SUFFIX`: a fixed reply string would be served from a prior run's cache on a rerun, bypassing usage tracking and leaving the budget permanently at zero spend.

## Verification

All eight pass against the live release stack on a fresh `QA_SUFFIX`, so they are rerunnable and order-independent as documented:

```
S197 0  S198 0  S199 0  S200 0  S201 0  S202 0  S203 0  S204 0
```

S199 and S201 were re-run after the assertion changes and still pass.

## Summary by CodeRabbit

- **Documentation**
  - Expanded release testing documentation with scenarios for budgets scoped to request labels.
  - Added coverage for label matching, multiple labels, managed API key labels, budget resets, validation errors, and authentication requirements.
  - Documented behavior across PostgreSQL and MongoDB backends.

- **Tests**
  - Increased the release scenario matrix from 196 to 204 scenarios.
  - Added enforcement and conflict-resolution checks for label-based budgets.

### DIFF
--- a/tests/e2e/release-e2e-scenarios.md
+++ b/tests/e2e/release-e2e-scenarios.md
@@ -465,6 +465,32 @@ run_release_budget_enforcement() {
     all(.budgets[]?; .user_path != $user_path)
   ' "$budget_json_file" >/dev/null
 }
+
+# Start a freshly created budget from zero spend.
+#
+# Spend is a SUM over usage rows, so deleting and recreating a budget does not
+# forget what an earlier run charged against the same subject. A scenario that
+# exits part-way on a failed assertion skips its own cleanup, and the rerun this
+# file recommends — same --qa-suffix, so the same labels — would then meet an
+# already-exhausted tiny budget and get 429 on its first request, failing for a
+# reason that has nothing to do with what it tests. Resetting after creation
+# makes that first request behave the same on the first run and the fifth.
+reset_release_budget() {
+  local base_url="$1"
+  local scope="$2"
+  local subject="$3"
+  local auth_header="${4:-}"
+
+  if [ -n "$auth_header" ]; then
+    curl -fsS -H "$auth_header" -X POST "$base_url/admin/budgets/reset-one" \
+      -H 'Content-Type: application/json' \
+      -d "{\"scope\":\"$scope\",\"subject\":\"$subject\",\"period\":\"daily\"}" >/dev/null
+  else
+    curl -fsS -X POST "$base_url/admin/budgets/reset-one" \
+      -H 'Content-Type: application/json' \
+      -d "{\"scope\":\"$scope\",\"subject\":\"$subject\",\"period\":\"daily\"}" >/dev/null
+  fi
+}
 ```
 
 ## Auth-enabled runtime environment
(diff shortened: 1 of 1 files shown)


SantiagoDePolonia:
Both addressed.

**Failed scenarios retain budget state (P1)** — confirmed and fixed. The key detail is that deleting and recreating a budget does *not* clear it: spend is a SUM over usage rows, so a fresh `PUT` inherits whatever the previous run charged against the same label. Cleanup at the end of a scenario can't help when the scenario is what failed.

New `reset_release_budget` helper resets each budget right after creation, so the first request behaves the same on the first run and the fifth. Verified by reproducing the exact situation — leave an exhausted budget behind, then rerun S198 with the same `--qa-suffix`:

```
leftover budget: {"spent":0.0000019,"remaining":-0.000001899}
without the reset:  S198 rerun exit: 1
with the reset:     S198 rerun exit: 0
```

Identical leftover state in both runs, so the reset is doing the work.

Worth noting the same exposure exists in the pre-existing `run_release_budget_enforcement` helper behind S87–S89 — it also cleans up only at the end. I left it alone to keep this PR scoped, but it would want the same treatment.

**Preamble overclaims** — correct, and it was wrong in both directions. Rewritten to say which scenarios configure tagging headers and restore the rule set (S198, S199, S201, S203), which uses managed-key labels instead (S200), and which are admin-only with no model traffic (S197, S204).

Full set re-run after both changes:

```
S197 0  S198 0  S199 0  S200 0  S201 0  S202 0  S203 0  S204 0
```

PR #608, opened 2026-07-28
fix(dashboard): stop live session-thread counts inflating on first-session requests

SantiagoDePolonia:
## Problem

With "Group by session" enabled and the Audit Logs page open, sending the first messages of a fresh session made the thread count badge (`audit-thread-count`) read 2x–3x the real number of requests. Refreshing the page fixed it, and later messages appended to an existing session counted correctly.

## Root cause

`audit.started` live events are published before session detection stamps the request context, so every live row is first inserted sessionless and only later events carry the `session_id`. When a later event made two on-screen rows collapse into one thread (or a new live request displaced a fetched head), the demoted row was **dropped** if the thread's children list had never been lazily fetched — `prependLiveAuditThreadChild` was a no-op for unloaded threads.

Every subsequent live event for that dropped request (updated, stream chunks, completed, flushed) then matched nothing on screen, so `foldLiveAuditIntoThread` re-counted it as a brand-new session member — +1 on the badge per event. The bug only shows on a session's first live-rendered messages because that is the only time several rows of one session are live with events still in flight.

## Fix

- `prependLiveAuditThreadChild` now always retains the displaced entry: for never-expanded threads it creates a **partial** children list marked `loaded: false`. Late events find their entry and merge in place instead of re-counting.
- `toggleThread` treats partial lists as unfetched (checks `loaded`/`loading` rather than slot existence), so expanding still fetches the full session page.
- `fetchThreadEntries` keeps live-displaced entries visible while the page loads and merges them on top of the fetched page via the new `mergeAuditThreadChildren` helper (deduped by id/request_id, including against a head replaced mid-flight; still-unpersisted live entries are preserved and counted into the thread total). Failed/stale fetches keep the partial entries instead of deleting the slot.

## Refactor (while there)

`mergeLiveAuditEntry` carried a near-complete duplicate of its merge flow for `audit.detail` events; it is now a single flow with an `isDetail` flag (detail events still never re-trigger the detail fetch, and never fold), with lifecycle-state stamping extracted into `liveAuditPatch`. `auditThreadChildEntries` grew into `mergeAuditThreadChildren` and the old helper was removed.

## User-visible impact

The thread badge now equals the number of distinct requests in the session while watching live traffic. Transient approximations remain only in genuinely unknowable cases (e.g. events replayed after a thread scrolled off the page) and reconcile on the next `/admin/audit/sessions` refetch, which was always correct.

## Tests

- New regression test replaying the exact reported sequence (two interleaved requests, sessionless `audit.started`, late session ids, completed/flushed for both): badge stays at 2 — the old code produced 4.
- New tests for displaced-entry retention and the fetch-merge helper; the test asserting the old dropping behavior was rewritten to the new invariant.
- All 385 dashboard tests pass, `svelte-check` clean, dist rebuilt.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audit-log session expansion to merge fetched results with live-displaced entries arriving mid-load, keeping totals accurate.
  * Prevented duplicate fetching by respecting already loading/loaded thread states.
  * Fixed session/thread counting issues during interleaved live events, avoiding inflated totals.
  * Ensured audit detail handling updates consistently without triggering premature expanded-detail fetches.
  * Retained displaced thread heads when folding into unexpanded threads instead of dropping them.

* **Tests**
  * Updated and added coverage for thread head/child merging, preserved-entry behavior, and session counting regressions.


SantiagoDePolonia:
Addressed both review findings in d142d307:

**Head replacement mid-flight (CodeRabbit major / Greptile P1) — confirmed, fixed as proposed.** Excluding the original `head` alongside `currentHead` was backwards: when a live request replaces the head during the session-page fetch, the original head is a demoted child that the fetched page must keep. `mergeAuditThreadChildren` now excludes only the current on-screen head (`currentHead ? [currentHead] : [head]`). When the head object was merely replaced by a live merge of the same request, the identity keys are identical, so behavior there is unchanged. Added a regression test for the race (`mergeAuditThreadChildren keeps a head demoted while the fetch was in flight`).

**Failed/stale fetch leaves a dead expanded thread (CodeRabbit minor) — valid, pre-existing before this PR, fixed.** When the restore path drops an empty slot, the thread is now also collapsed, so a single click retries instead of three. When live-displaced entries remain, the thread stays expanded showing them, and the next expand still refetches (`loaded: false`).

PR #611, opened 2026-07-28
feat(ext): add route selector extension point and adaptive strategy

SantiagoDePolonia:
## Summary

Adds a routing extension point to the public `ext` API so extension builds can steer load balancing for virtual models:

- **`ext.RouteSelector`** (`ext/route.go`): `Select` picks among currently viable targets of a load-balanced virtual model; `OnAttemptStart`/`OnAttemptEnd` observe every upstream attempt (primaries, retries, failovers) via the llmclient hooks, so selectors score targets from real traffic. Registered through a single slot on the registry (`RegisterRouteSelector`).
- **New `adaptive` load-balancing strategy** for virtual models: delegates target choice to the registered selector. Candidates carry provider/model, configured weight, and registry pricing.
- **Wiring** (`internal/app/app.go`): selector hooks attach before providers are created (same rule as the health tracker); the virtual-models service gets the selector alongside the capacity probe.

## Safety / portability

- Session affinity, rate-limit capacity filtering, failover chains, and retries all stay in core — the selector only picks the primary among viable targets.
- Fail-open by design: no selector registered, selector declines, answers outside the pool, or **panics** (recovered and logged, both in selection and observation) → exact `round_robin` behavior. Configs using `adaptive` stay portable between plain core and extended builds.
- Single-viable-target pools bypass the selector, so aliases behave identically with and without one.
- An empty registry keeps adding zero request overhead.

## Tests

- `ext`: registry single-slot semantics, `RouteTarget.Qualified()`.
- `virtualmodels`: selector delegation (request shape incl. pricing), fallback matrix (no selector / decline / out-of-pool answer / panic), single-target bypass.
- Full suite + race detector green (pre-commit `make test-race`).

## Summary by CodeRabbit

* **New Features**
  * Added **adaptive** virtual-model load-balancing with selector-driven target selection and fallback to weighted round-robin.
  * Enabled route-selector extensions to steer routing using session/candidate context and attempt lifecycle callbacks.
  * Updated the dashboard/editor to surface available strategy options from runtime configuration.
* **Documentation**
  * Expanded config and help text to document the **adaptive** strategy and its “extension required, otherwise fallback” behavior.
* **Tests**
  * Added coverage for selector registration behavior, **adaptive** delegation/fallback (including panic/decline), single-target bypass, candidate pricing copying, and route target formatting.


SantiagoDePolonia:
Addressed the review findings in 189beca5:

- **Recovery re-panic (P1)**: the selector's name is now captured panic-safe at install time (`selectorLabel`, both in `virtualmodels` and `app`); recovery handlers log only fixed metadata and never call back into extension code. The raw panic value is no longer logged (extension-controlled, could carry request data). Regression test added for the Select-and-Name-both-panic path.
- **Per-attempt contract (P1)**: fixed the contract, not the plumbing — `ext/route.go` now documents that hooks fire once per upstream call, with transport-level retries inside the provider client aggregated into that call's outcome (final status; duration spans retry backoff, so retry-heavy targets still score slower). Moving observation inside the client retry loop would change `llmclient.Hooks` semantics for every existing consumer (Prometheus metrics, health tracker) and belongs in its own PR if we ever want per-transport-attempt observability.
- **Pricing escape (P2)**: candidates now carry defensive copies of the catalog's per-Mtok prices; test asserts a selector writing through the pointers cannot change what cost routing later reads.
- CodeRabbit: example-config wording fixed (adaptive is optional, falls back to round_robin), registry selector-slot tests are table-driven (unset/single/replace/nil-reset), and the delegation test asserts the complete candidate set.

#612 has been rebased on top.

SantiagoDePolonia:
Follow-up from the second review pass in b3c07a4b: the pricing-isolation test now mutates every pricing pointer of every candidate and verifies all catalog prices stay intact.

SantiagoDePolonia:
Added in 454b7cc2 (discussed with the team): the strategy dropdown is now **server-driven**. `/admin/runtime/config` carries a new `VIRTUAL_MODEL_STRATEGIES` key (comma-separated); `adaptive` is listed only when a route-selector extension is registered, so plain-core dashboards never offer a strategy that would silently fall back to round robin, while extended builds get it configurable from the UI. Values come from the backend, labels stay in the frontend (unknown values render raw for forward-compat), and the editor keeps an edited row's current strategy selectable even when the deployment doesn't offer it — fixing the pre-existing blank-select trap when opening an API-created adaptive VM on plain core. Covered by the existing runtime-config contract test plus new Go and dashboard unit tests; dist rebuilt.

SantiagoDePolonia:
Addressed in fe6a0e60: the adaptive label and help text now describe the contract ("target chosen by the registered routing extension") rather than one extension's policy.

SantiagoDePolonia:
CodeRabbit's help-text finding fixed in 094d1d18: the adaptive help sentence now gates on the server-provided strategy list (`runtimeConfig.virtualModelStrategies()`) instead of the dropdown options, which include an edited row's unsupported value for compatibility — so an adaptive VM opened on plain core no longer shows help claiming an extension picks targets. #612 merged forward.

PR #619, opened 2026-07-30
perf(dashboard): speed up the audit logs page

SantiagoDePolonia:
The Audit Logs page felt slow and expanding a record looked like it animated twice. Both turned out to be real, and measurable.

Everything below was measured in a headless Chrome driving the actual page, and with `EXPLAIN QUERY PLAN` plus timings against a real 27,139-entry / 16,979-thread database — not estimated.

## The double animation

`transition:slide` measures its target height **once, when the transition starts**. Opening a row also kicks off the `/admin/audit/detail` fetch, so the real bodies land mid-animation:

```
before   40 → 371px over 150ms,  then an instant snap to 593px   ← two motions
after   373 → 595px, one monotonic ease-out, no snap
```

The panel now fades and lifts (`transition:fly`) instead of sliding its height. That is immune to late-arriving content and stays off the layout path entirely — no per-frame relayout of everything below the row.

Trade-off worth knowing: on close the panel now fades at full height and the row collapses at the end, rather than sliding shut. The close never had the stale-height bug (its content is settled), so `in:fly` + `out:slide` would keep both behaviours if that reads better.

## Redundant work on the page

| | |
|---|---|
| `motionDuration()` allocated a `MediaQueryList` per call | ~50 per list render → created once |
| `AuditAttemptTrack` rebuilt + re-sorted the attempt list 5× per row (`auditAttemptTrackTitle` twice, for `title` and `aria-label`) | derived once |
| `auditEffectiveTab` rebuilt the whole pane set to validate one tab id | takes the panes the caller already holds |
| fetched rows carried a zero-duration `slide`, whose keyframes still cost a `getComputedStyle()` each | paging: **50 → 25** style recalcs (measured) |
| `fetchAuditLog` flattened every loaded thread's children twice | hoisted |
| `.live-dot.is-streaming` pulsed via `box-shadow` — repaints every frame, forever, per streaming row | pseudo-element animating `transform`/`opacity` |

## Deduplication

`formatJSON` was duplicated byte-for-byte across `audit-logic.js` and `conversation-helpers.js`, as was `tryParseJSON`. The two modules also carried **separate nested-error walkers** — one recursive at depth 6 that unwrapped JSON-in-strings, one breadth-first at depth 4 that skipped strings — so the same entry could show different error text in the pane and in the Interactions drawer.

One shared implementation in `error-text.js` now serves both, with the drawer's `extractText` fallback preserved as an optional parameter. **User-visible consequence:** the drawer now unwraps JSON-string response bodies the way the pane already did, so both finally agree. Net **-247/+141** lines.

## `GET /admin/audit/sessions`: 351ms → 128ms

Grouped-by-session is the default view, so this is paid on every page load, filter change and date change. The flat `/admin/audit/log` served the same window in 2.4ms.

Both readers ranked threads while carrying whole entries — the SQL window pass selected every `logColumn` including the `data` blob, and MongoDB sorted full documents and kept one per thread in group state via `$first: "$$ROOT"`. All of that to return 25 rows totalling 37KB.

Both now rank over **id and timestamp only** and re-read the page's heads afterwards (by primary key in SQL, by `_id` in a second round trip for Mongo). The SQL window functions additionally share one named window, so the partition sorts once rather than once per aggregate, and the ranked row's own timestamp replaces the `MAX()` window — the partition is ordered `timestamp DESC`, so `rn = 1` *is* the thread's newest entry. The explicit full frame is what lets the aggregates reuse the ranking window's `ORDER BY` without becoming running totals.

| | before | after |
|---|---|---|
| page query | 204 ms | **49 ms** |
| count query | 24 ms | 24 ms (untouched) |
| endpoint | 351 ms | **128 ms** |

Plan drops from three temp b-tree sorts to two.

### Deliberately not done

- **The 24ms `COUNT(DISTINCT …)`.** It folds into the page query as `COUNT(*) OVER ()`, but an offset past the end then returns zero rows and a silent total of 0. Not worth a correctness edge case for ~15% of what remains.
- **An expression index on the thread key.** Tried it on a copy of a real database: the planner ignores it, because the `WHERE` is a timestamp range and only one index can serve. Zero gain, permanent write cost.
- **A `GROUP BY` + correlated-lookup rewrite.** Benchmarked at 200ms — the per-thread lookup cannot use an index on the thread-key expression, so each of the 25 lookups rescans the window.

## Verification

- Old vs new endpoint output over **16,767 real threads**: `JSON.stringify(a) === JSON.stringify(b)` → true. Query results also diffed directly over 200 rows at a non-zero offset: byte-identical.
- `TestSQLReader_GetSessions` run against **SQLite, PostgreSQL 18 and MongoDB 8**, including `TestSQLReader_GetSessionsOnLegacyUUIDSchema` on the pre-unification uuid `id` column — the case the `CAST(id AS TEXT)` exists for.
- Two shared assertions added, covering exactly what the rewrite risks: `assertGetSessionsHeadPayload` (the head still carries its list columns *and* its `data` payload — the old fixtures had no `Data`, so a broken re-read would have gone unnoticed) and `assertGetSessionsPaging`.
- `go test ./...`, `make lint` (0 issues), dashboard `npm test` (406/406), `svelte-check` (0 errors).

## Not addressed

Three findings from the same investigation are left for follow-ups, since each is a larger change:

1. **`animate:flip` replays the whole list per live event.** Svelte calls `measure()`+`apply()` on every keyed item on every each-block update. Measured with 25 rows while firing 6 gateway requests: **700 `getBoundingClientRect()` and 516 `element.animate()` calls** — ~117 forced layout reads per request. Root cause is `live-logs-logic.js` replacing the entries array identity on nearly every merge, including field-only updates.
2. **Hidden tab panels render their bodies.** Every `AuditPane` stays mounted, so a 3-attempt entry escapes and `{@html}`s four full JSON bodies, three invisible — re-run several times a second during a streamed response.
3. **`AuditThreadGroup` still uses `transition:slide`** and has the same stale-height shape as the row expansion did.


SantiagoDePolonia:
Addressed the Greptile P1 in dd0b2656.

The finding is correct and it was a regression this PR introduced. The narrow grouping pass fetched the page's full documents in a **second round trip**, so an entry deleted by retention between the aggregation and the `Find` was silently skipped — a short page while `total` still counted it. The default view is the last 30 days, which is exactly where the retention boundary sits, so the last page was the realistic victim.

Worth noting the SQL reader never had this: its `heads` CTE and the join back are one statement, so they see a single snapshot. Only MongoDB was exposed, because only MongoDB needed two queries.

Fixed by moving the head fetch into a `$lookup` inside the `$facet`, after `$skip`/`$limit`:

```
$match → $project{_id, session_id, timestamp} → $sort → $group → $sort
       → $facet{ data: [$skip, $limit, $lookup, $unwind], total: [$count] }
```

The heads are now read in the same pipeline that selected them, and the round trip is gone. The `$project` before `$sort` still keeps whole documents — bodies included — out of the sort and out of the group state, which was the point of the original change. `logEntriesByID` is deleted; the decode goes back to `mongoLogRow`, so the diff is now smaller than before.

Verified against MongoDB 8 (`MONGO_TEST_DSN`), including `assertGetSessionsHeadPayload`, which asserts the head still carries its `data` payload — the precise thing the `$lookup` has to re-attach.

---

On the Codecov 0%-patch-coverage flag for this file: `internal/storage/mongotest` skips unless `MONGO_TEST_DSN` names a reachable server, and CI does not set it, so no MongoDB reader code is covered in CI — pre-existing and by design per that package's doc comment, not specific to this change. Greptile's T-Rex hit the same wall (no Docker in its sandbox). I ran the suite locally against MongoDB 8, plus PostgreSQL 18 for the SQL side including the legacy uuid-id schema.

PR #620, opened 2026-07-31
fix(dashboard): keep live audit inserts flowing for today-anchored ranges; silence unload stream errors

SantiagoDePolonia:
## Problem

Two independent issues around the live audit-log stream, found while debugging "I can't see the realtime logs":

**1. Live inserts silently disabled by a persisted date range.** Since #617 the reporting window persists across sessions. A custom range re-applies `customStartDate`/`customEndDate` on every load, and `auditLiveInsertAllowed()` pauses live inserts whenever those are set — so one custom range selection disables live logs forever, with a connected, delivering stream and no UI hint. Verified end to end: the SSE stream was healthy (server heartbeats, events delivered in headless Chromium and Firefox), while the page dropped every event at the merge gate because localStorage held `{"mode":"custom","start":"2026-07-16","end":"2026-07-31","follow":true}`.

**2. One console error per refresh in Firefox.** Page unload kills the in-flight stream fetch, and Firefox rejects it with plain TypeErrors (`Error in input stream` mid-read, `NetworkError when attempting to fetch resource` pre-headers) instead of an `AbortError`, so `isAbortError()` misses them and every refresh logs `Live logs stream failed: TypeError: …`. Reproduced via Playwright Firefox: `NS_BINDING_ABORTED` on the stream request at each reload.

## Fix

- `auditLiveInsertAllowed()` lets a custom range that **ends today** (`followsToday`, already tracked by the date-range store) keep live inserts flowing — an entry arriving now is inside such a window by definition. Detached historical ranges still pause inserts, as before.
- The stream singleton aborts itself on `pagehide` (teardown becomes a deliberate abort), the reader trusts `controller.signal.aborted` over the error's name, and a back/forward-cache restore (`pageshow` with `persisted`) restarts the stream if it was running.

## Testing

- Two new gate cases in `tests/live-logs.test.js` (today-anchored range inserts; detached historical range still blocked); full dashboard suite: 411 pass.
- Dist rebuilt via `make frontend`.

## Summary by CodeRabbit

* **New Features**
  * Added live audit-log status indicators for connecting, active, and paused states, with guidance when filters or date ranges pause updates.
  * Live-log streams now recover when returning to a previously open page.
  * Live subscriptions provide an immediate heartbeat for faster connection feedback.
* **Bug Fixes**
  * Live audit logs continue updating for custom date ranges ending today.
  * Improved handling of intentionally stopped streams.
* **Changes**
  * Moved session grouping to the audit-list toolbar alongside the live status indicator.
  * Removed session grouping from the audit filter controls.


SantiagoDePolonia:
Added in `4d2f73c6` (scope grew from review of the fix):

- **Live-status indicator** on the audit list, right of the "Showing X–Y of Z" summary: pulsing green dot + **Live** while inserts flow; muted dot + a hint naming the actual blocker when paused ("Live paused — the selected date range does not include today. Set it to today to resume." / filters / later page). Hidden when `DASHBOARD_LIVE_LOGS_ENABLED` is off. Reuses the existing `.live-dot` pattern from Live Tokens.
- **"Group by session" moved** from the filter grid to sit next to the summary it describes.
- **Immediate SSE heartbeat on subscribe** (`handler_live.go`): previously nothing reached the wire until the first ticker heartbeat, so response headers sat buffered for up to 15s and a fresh stream looked pending the whole interval.

Verified in headless Firefox against a live build: indicator turns Live within ~2s of load, a triggered event inserts at the top of the list, a detached range shows the paused hint, and a today-anchored range stays Live. JS suite 413 pass; admin Go suite green.

SantiagoDePolonia:
All review findings addressed in `bda3bf33`:

- **Greptile P2 — reconnect after pagehide stop** (`liveLogs.svelte.js`): valid. A reader that resolved `done` normally right after `stopLiveLogs()` unconditionally scheduled a reconnect, opening a stream nothing owns (the harness reproduced it). The resume path now returns early when the read's own controller was aborted or replaced. This also covers the pre-existing SPA-teardown variant of the same race.
- **CodeRabbit Major — heartbeat sequence race** (`handler_live.go`): valid, and it ran deeper than the initial heartbeat — ticker heartbeats and the reset event had the same flaw, pre-dating this PR. `Subscription` now captures `Latest` atomically with the replay snapshot (under the broker mutex), and the handler tracks the newest sequence actually written to the connection (advancing past deliberately-filtered events), advertising that as every heartbeat's cursor. Broker test pins the snapshot semantics: events published after subscribe stay off `Latest`.
- **CodeRabbit Minor — weak test assertion**: fixed; the regression test is now table-driven per repo guidelines and asserts `id: 1\nevent: heartbeat\n` as one frame, plus a fresh-broker case asserting the seq-0 heartbeat omits the SSE id line.
- **CodeRabbit Minor — `.live-dot` styles missing in component**: not valid — the base `.live-dot`/`.is-streaming` styles are global (`cards-charts.css`, same pattern as LiveTokens and AuditPane) and bundle into the single dist stylesheet; rendering verified in headless Firefox screenshots. The component adds only its scoped throb/spacing on top.
- **CodeRabbit Minor — summary margin override**: already fixed in `57f57654` (the stale duplicate rule was removed; the universal reset supplies the zero margin).

Suites: internal/live + internal/admin green, dashboard JS 413 pass, dist rebuilt; verified live against a running build (immediate heartbeat on connect, live insert renders, indicator states correct).

PR #644, opened 2026-08-03
feat(server): reload configuration without a restart via --reload

### LINKED ISSUE #573: config refresh mechanism without involving app restart like - nginx reload
@lamualfa said: "...it would be great if there is config refresh mechanism without involving app restart like nginx reload."

source: https://github.com/ENTERPILOT/GoModel/issues/555

SantiagoDePolonia:
`gomodel --reload` now asks the running gateway to re-read its
configuration, the same operation as `nginx -s reload`. It resolves the
new `server.pid_file` / `PID_FILE` setting, signals that process
(SIGHUP, so `kill -HUP` works too), and exits.

On the signal the gateway re-reads the `.env` file and then reloads the
whole configuration by rebuilding itself, so every setting reloads
rather than a curated subset. Two properties make it safe to run in
production:

- the replacement is built before the running one is stopped, so a
  configuration that fails to load or initialize leaves the gateway
  serving on the one that already works;
- the listening socket is owned by the process and handed to each
  generation in turn, so requests arriving mid-reload wait in the accept
  queue instead of being refused.

Environment file handling keeps startup's precedence: variables
exported into the process still win over the file, edited values are
applied, and variables removed from the file are unset.

`PORT` and `PID_FILE` changes still need a restart and are warned
about. The pid file defaults to `data/gomodel.pid` next to an existing
`./data` directory and to the OS per-user data directory otherwise —
the same resolution the SQLite database uses.

Closes #573

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01F5FhrZ1ZchNHmEnvsTXdty

## Summary by CodeRabbit

* **New Features**
  * Added `gomodel --reload` to apply environment and configuration changes without stopping the gateway.
  * Listening sockets remain available during successful reloads.
  * Added configurable PID-file support, including per-instance paths and an option to disable PID-file creation.
  * Reload failures preserve the currently running gateway and active configuration.
  * Reload is available on POSIX platforms and reports signaling or configuration errors.
  * Port and PID-file changes require a restart.

* **Documentation**
  * Added guidance covering reload behavior, PID files, limitations, and restart-required settings.


SantiagoDePolonia:
Review findings addressed in `84c1add` and `0e35bbd`.

**Pre-bound listener bypassed the gateway HTTP server configuration** (both reviewers, P1) — real, and already fixed in `84c1add` before these reviews landed; they were run against `a6c7bfa`. `Server.StartWithListener` built a bare `echo.StartConfig`, so moving production serving onto the listener path dropped the inbound read/header/write timeouts and the gateway's graceful drain window. Both start paths now go through `newGatewayStartConfig`, with a test on the listener variant.

**Failed reload leaked the new environment and logger into the retained generation** (both reviewers, P1) — real, fixed in `0e35bbd`. `rebuild` has to read `.env` and install the new logging configuration before it can build the replacement (config loading reads the process environment, and the new log level should cover the build itself), but both are process-wide. Both are now rolled back when the build fails, including the dotenv bookkeeping, so a later reload still applies the same file. Verified end-to-end: after a rejected reload the retained generation emits no DEBUG output and keeps its master key; the next successful reload applies both.

**Register the reload signal before writing the pid file** — real. The pid file is what advertises that this instance can be signalled, and until `Notify` runs, SIGHUP still terminates the process. Reordered.

**`PID_FILE` restart-required, in the places an operator actually reads** — done in the struct comment, `.env.template`, `config.example.yaml`, `configuration.mdx`, and `config-yaml.mdx` (the latter for `server.port` too).

**Empty `PID_FILE` case** — worth adding, because it failed: empty env vars read as unset throughout this config, so `PID_FILE=` kept the default rather than disabling the pid file. The docs claiming otherwise were wrong; disabling is `server.pid_file: ""` in `config.yaml`. `TestPIDFilePathResolution` now pins all three outcomes.

**No-refusal guarantee depends on descriptor duplication** — qualified in `cli.mdx`, matching the code comment.

**Extract the shared data-directory fallback** — done: `platformdir.DataFile`, called by both `DefaultSQLitePath` and `DefaultPIDFilePath`. The local form keeps its forward slash so the legacy path constants stay comparable on every platform.

**Coverage for `sendReloadSignal`** — added, table-driven: signal delivery to a `signal.Notify` channel, missing pid file, and a pid file naming no process. Also added `ReadTimeout` and the nil-listener error case to the listener start-config test.

Full suite, `-race`, e2e, and contract tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #672, opened 2026-08-10
feat(elevenlabs): add ElevenLabs voice provider

SantiagoDePolonia:
## Summary
- Add a new `elevenlabs` provider type: a voice-only provider exposing OpenAI-compatible `/v1/audio/speech` (text-to-speech) and `/v1/audio/transcriptions` (speech-to-text). Chat, `/v1/responses`, and embeddings return `invalid_request_error` since ElevenLabs has no such APIs.
- The OpenAI `voice` field carries the ElevenLabs `voice_id` directly (ElevenLabs has no named voices); auth uses the `xi-api-key` header rather than Bearer.
- `ListModels` merges ElevenLabs' live TTS catalog (`GET /v1/models`, filtered to `can_do_text_to_speech`) with the fixed speech-to-text model list (`scribe_v2`, `scribe_v1`), which is not included in that listing.
- Speech supports `mp3`/`opus`/`pcm`/`wav` response formats and `speed` 0.7–1.2 (ElevenLabs' voice-setting range); transcription supports `json`/`text`/`verbose_json` with word-level timestamps mapped from ElevenLabs' `words` array.
- Wired into `run/providers.go`, dashboard help-icon doc link, `.env.template`, `config/config.example.yaml`, README, CLAUDE.md, and a new dedicated `docs/providers/elevenlabs.mdx` guide (added to nav).
- Request/response shapes were cross-checked against ElevenLabs' official Python SDK type definitions (Fern-generated, source of truth) rather than relying solely on prose docs.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (full suite green, including new `internal/providers/elevenlabs` package tests)
- [x] `go vet ./internal/providers/elevenlabs/...`
- [x] Dashboard JS unit tests (`node --test tests/*.test.js`, 477 passing)
- [x] Pre-commit hooks (fmt, lint, dashboard dist sync, mint validate) all pass

## Summary by CodeRabbit

* **New Features**
  * Added ElevenLabs support for text-to-speech and speech-to-text.
  * Supports MP3 and WAV audio, transcription formats, timestamps, voice IDs, and model discovery.
  * Added API key and optional custom base URL configuration.
  * Added passthrough support for native ElevenLabs requests.
  * Unsupported chat, responses, and embeddings operations now return clear errors.

* **Documentation**
  * Added setup guidance, capability details, configuration examples, and provider documentation links.


SantiagoDePolonia:
Reviewed and verified each finding against the actual code/behavior; pushed fixes for the valid ones.

**Fixed:**
- **Greptile (P1)** — Static Scribe transcription models could disappear if the very first `GET /v1/models` fetch fails (cold start), since `ListModels` returned early before appending them. Fixed with a narrower fallback: on a fetch failure with no prior successful fetch, return the static models; once a fetch has succeeded, later failures propagate normally so `internal/providers/registry_provider_refresh.go`'s existing stale-inventory carry-forward keeps the larger prior list — returning the static-only list unconditionally on every failure would have been a regression there (it would silently shrink an already-populated catalog instead of leaving it stale-but-intact).
- **CodeRabbit** — `speechSpeed` now clamps to ElevenLabs' 0.7–1.2 range instead of rejecting values OpenAI clients legitimately send (0.25–4.0), matching this repo's Postel's Law principle (and the same clamping pattern `minimax`'s temperature handling already uses).
- **CodeRabbit** — Fixed a stale comment claiming the provider doesn't implement `core.PassthroughProvider`, when it does.
- **CodeRabbit** — Added the missing test coverage it flagged (pcm/word-granularity mapping, wav format, speed clamping, table-driven unsupported-capability checks, an `elevenlabs` row in `TestDefaultProviderFactoryCredentialForms`), plus new tests for the catalog-fallback fix above.

**Skipped after verification (false positives):**
- **CodeRabbit** — "buffer the multipart transcription body instead of streaming through `io.Pipe`, it's not retry-safe." `llmclient.DoRaw` already forces `maxAttempts = 1` whenever `RawBodyReader` is set, and calls `closeRawBodyReader` (which closes the pipe reader, unblocking the writer goroutine) on every early-exit path. This is the identical pattern `cohere/audio.go`'s `cohereTranscriptionMultipart` already uses in this codebase.
- **CodeRabbit** — "use `opts.Keys` instead of `opts.Keyring(cfg.APIKey)`, so factory-created providers use only the primary key." `ProviderOptions.Keyring()` already returns `opts.Keys` when the factory set it, only falling back to a single-key ring when constructed outside the factory (tests, `NewWithHTTPClient`). Same pattern `cohere.go` already uses.

Also independently double-checked the original implementation against ElevenLabs' actual API surface (via their Fern-generated SDK source, not just prose docs) before this review pass and found two real bugs in the original PR body, already fixed in an earlier commit: `wav` was wrongly rejected as an unsupported speech format, and the static transcription model list had a nonexistent `scribe_v1_experimental` while missing the current `scribe_v2`.

SantiagoDePolonia:
## E2E test results against the live ElevenLabs API

Ran the gateway locally with a real `ELEVENLABS_API_KEY` and exercised TTS, STT, and passthrough extensively (not mocked). Full round-trip worked: synthesized "The quick brown fox jumps over the lazy dog." → transcribed it back → got the same text. Also tested a ~40s / 625KB longer synthesis → transcription (verbose_json, 85 words, accurate timestamps), all 4 speech formats, all 6 TTS models, error paths, and speed clamping (confirmed the clamped value actually changes synthesized audio duration, not just accepted silently).

**Found and fixed one real bug** (pushed): `CreateSpeech`/`CreateTranscription` had a status-code check that was dead code — `llmclient.Client.DoRaw` already converts any non-2xx response into an `error` before returning, so `resp.StatusCode` can only ever be 200 by the time my check ran. This meant ElevenLabs' actual error detail (`{"detail": "..."}` or `{"detail": {"message": "...", ...}}`) was never unwrapped; the generic client-level parser only recognizes `{"message": ...}`/`{"error": {...}}` and fell back to dumping the raw JSON body as the error message, e.g. `"message": "{\"detail\":{\"message\":\"bad key\",...}}"` instead of `"message": "bad key"`. Fixed with `refineElevenLabsError`, which post-processes the `*core.GatewayError` DoRaw already built (using its preserved `ResponseBody`). Verified against live 400/401/403 responses — messages are now clean.

**Found a real bug outside this PR's scope** (documented, not fixed here): the provider-passthrough router's `ALLOW_PASSTHROUGH_V1_ALIAS` handling in `internal/server/passthrough_support.go` strips a leading `v1/` path segment for *every* provider uniformly, assuming the provider's own base URL already embeds `/v1` (true for OpenAI-shaped providers, which is presumably why this was never caught). ElevenLabs' base URL is `https://api.elevenlabs.io` with no `/v1`, so every `/p/elevenlabs/v1/...` passthrough call 404s — verified live (`/p/elevenlabs/v1/voices` → 404, `/p/elevenlabs/voices` → also 404, `/p/elevenlabs/v2/voices` → 200, confirming the mechanism precisely). Practical impact: speech-to-speech (voice changer) is completely unreachable through GoModel today since it only has a `/v1` path with no `/v2` alternative to route around the bug with. I documented this (mechanism + impact + workaround) in `docs/providers/elevenlabs.mdx` under a new "Not implemented" section, since fixing it properly means teaching the passthrough router which providers' base URLs already embed `/v1` — a small but genuinely cross-cutting change affecting every provider's passthrough routing, not something I want to drive-by in an ElevenLabs-scoped PR. Happy to open a follow-up for it if useful.

Also noted (not fixed, pre-existing, confirmed via `internal/server/audio_service.go`): audit log entries for `/v1/audio/speech` and `/v1/audio/transcriptions` don't populate the top-level `provider` field for any audio-capable provider, not just ElevenLabs — `requested_model` is populated correctly, just not `provider`.

PR #694, opened 2026-08-16
perf(server): cut hot-path overhead and guard the production-shaped path

SantiagoDePolonia:
## Summary

Two-part change: measured hot-path quick wins, and a perf-guard extension so CI can actually see the configuration deployments run.

### Part 1 — quick wins (`perf(server)` commit)

| Guarded benchmark | Before | After |
|---|---|---|
| Bare hot path | 6,814 ns / 110 allocs | **6,014 ns / 106 allocs (−12%)** |
| Routed hot path | 7,771 ns / 130 allocs | **7,435 ns / 126 allocs** |
| Production shape (new) | 18.5 µs / 294 allocs | **16.5 µs / 285 allocs** |

- **goccy response serializer**: request decode already uses goccy via the core types; every response still paid echo's reflection-based `encoding/json` encoder. Deserialize keeps echo's default (admin `c.Bind` error mapping untouched).
- **Single `ensureRequestID`**: the dedicated request-ID middleware repeated work `RequestSnapshotCapture` (unconditional, runs right after) already does — a second context wrap + request copy per request for no effect.
- **Session-anchor canonicalization**: goccy decode + `Decoder.More` instead of a full second trailing-data decode, ~5–12× per request. Canonical bytes are **byte-identical** to `encoding/json`'s — pinned by `TestCanonicalSegmentMatchesStdlib` — so auto-detected session ids do not re-anchor on upgrade (virtual-model affinity pins and downstream per-session consumers are unaffected).
- **Known-field set lookup** for the struct-derived lists (~40 entries scanned per JSON key before); hand-listed 2–5-field callers keep the linear scan, which wins at that size.

### Part 2 — production-shape guard (`test(perf)` commit)

The guard built the server with every default-on subsystem disabled (no auth, audit, usage, session keeping, rate limits), so nothing added to the middleware chain since it was written was visible to CI — the real default deployment costs ~2.2× the guarded number.

- `BenchmarkGatewayHotPathProductionShape`: auth + audit (bodies/headers) + usage + session keeping + a real `ratelimit.Service` with one configured rule, now **ceiling-enforced in `TestHotPathPerfGuard`** (runs in the existing Performance Guard CI job via `make perf-check` — no workflow change needed).
- `BenchmarkAblation*`: per-subsystem cost attribution against the full shape (diagnostic, `make perf-bench`).
- `TestSessionIDVisibilityByBodySize`: pins that content auto-detection is independent of body size, including past the 64 KiB snapshot inline-capture limit.

## User-visible impact

Lower per-request latency/allocations on every translated endpoint; no API or behavior change. Session ids, response bytes, and error shapes are unchanged (each pinned by test).

## Testing

Full suite (`test-race`), lint, and the extended perf guard pass locally (pre-commit runs all three); ceilings tightened to the new baselines.

Related: ENTERPILOT/GoModel-pro#20 removes the Pro-side same-session serialization this benchmarking uncovered.

## Summary by CodeRabbit

* **Performance**
  * Improved processing efficiency for chat and response payloads.
  * Optimized response encoding, including support for formatted output.
  * Improved request handling and correlation across server middleware.

* **Reliability**
  * Strengthened JSON compatibility and trailing-data validation.
  * Preserved consistent handling of Unicode, nested, numeric, and HTML-sensitive JSON content.

* **Testing & Documentation**
  * Added production-shaped performance benchmarks and subsystem comparisons.
  * Documented benchmark coverage and validated session detection for large request bodies.

### DIFF
--- a/internal/core/chat_json.go
+++ b/internal/core/chat_json.go
@@ -4,7 +4,7 @@ import "github.com/goccy/go-json"
 
 // chatRequestFields is derived from the struct's json tags at package init so
 // the known-field list cannot drift from the type definition.
-var chatRequestFields = jsonFieldNames(ChatRequest{})
+var chatRequestFields = jsonFieldSetOf(ChatRequest{})
 
 // UnmarshalJSON decodes the typed fields via an alias (so new fields are
 // picked up automatically) and captures every other member in ExtraFields.
@@ -15,7 +15,7 @@ func (r *ChatRequest) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	extraFields, err := extractUnknownJSONFields(data, chatRequestFields...)
+	extraFields, err := extractUnknownJSONFieldsSet(data, chatRequestFields)
 	if err != nil {
 		return err
 	}
--- a/internal/core/json_fields.go
+++ b/internal/core/json_fields.go
@@ -292,6 +292,24 @@ func (fields UnknownJSONFields) Without(keys ...string) UnknownJSONFields {
 	return UnknownJSONFields{raw: buf.Bytes()}
 }
 
+// jsonFieldSet answers "is this key a known typed field?" in O(1). Use it via
+// jsonFieldSetOf for the derived struct lists (dozens of fields, checked once
+// per JSON key of every decoded request); the variadic
+// extractUnknownJSONFields stays optimal for the hand-listed callers with a
+// handful of fields, where a linear scan beats a map lookup.
+type jsonFieldSet map[string]struct{}
+
+// jsonFieldSetOf derives the known-field set from v's struct definition, like
+// jsonFieldNames but as a lookup set.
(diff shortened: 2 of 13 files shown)


SantiagoDePolonia:
Both review findings addressed in 10bbaac2:

- **canonicalSegment trailing-data guard** (CodeRabbit): confirmed — goccy `Decoder.More()` returns false for stray closing brackets (`1]`, `1}`), so those would have been canonicalized instead of falling back to exact raw bytes. Restored the decode-to-`io.EOF` check; for valid input the extra decode reads only the empty remainder, so the hot path is unaffected (guard ceilings unchanged, all passing). Added both bracket cases to `TestCanonicalSegmentFallsBackToExactRawJSON`.
- **missing assertion** (Greptile): `TestSessionIDVisibilityByBodySize` now fails when any size stops producing a session id, instead of only logging it.

PR #699, opened 2026-08-17
feat(dashboard): fill loading-state gaps and speed up audit log search

SantiagoDePolonia:
### Loading states (dashboard)
- **Overview**: `SummaryCards` shows spinners in every stat value while the summary/cache fetches are in flight; `CacheMeter` shows a spinner on first load instead of flashing "No usage data"; `AuditStatsCharts` holds its place with a `LoadingState` section on first load instead of popping in.
- **Usage**: the cache stat cards get spinners, backed by a new `usageData.cacheLoading` flag — `fetchCacheOverview` previously tracked no loading state.
- **Search feedback**: `FilterInput` gains a `loading` prop (small trailing spinner inside the input), wired to the audit-log and usage request-log searches. Both searches were already debounced (300 ms).
- **DRY**: five hand-rolled centered-spinner blocks (audit logs, usage request log, API keys, workflows, guardrails) replaced with the shared `LoadingState` molecule; their bespoke CSS deleted.

### Audit log search (backend)
- **UUID fast path**: a pasted full canonical UUID (request id, API key id, session id — the most common expensive search) now matches the indexed identity columns by equality (case-insensitively, trying both spellings) instead of running `LIKE '%…%'` across 11 columns. Mirrored in the Mongo reader. Deliberate trade, documented in code: a UUID that only appears inside free text (e.g. an error message) no longer matches.
- **Gated JSON clause**: the `error_message` search term forced a JSON parse of the full `data` blob for every row in the date window. All writers set `error_type` whenever they write an error message, so the parse is now gated behind a non-empty `error_type` and healthy rows never pay it.
- **No new indexes needed**: every identity column searched is already indexed; leading-wildcard LIKE can't use indexes regardless, which is why the fast path switches to equality. Merging the separate `COUNT(*)` via `COUNT(*) OVER ()` was considered and rejected — it would defeat the page query's timestamp-index early termination by materializing every matching row.

## User-visible impact
- Every data-loading widget on the Overview and Usage pages now shows a spinner until its data arrives; search boxes show an in-flight spinner while typing triggers refetches.
- Searching a pasted UUID on the Audit Logs page is now an index lookup instead of a full scan; free-text searches skip the per-row JSON parse on non-error rows. Semantics change only for full UUIDs buried in error-message text.

## Testing
- New Go tests for the UUID fast path (uppercase paste, request-id/session-id hits, free-text exclusion), run green on both SQLite and a live PostgreSQL.
- `go test ./...`, `npm run check` (0 errors), `npm test` (495 pass), `npm run build` (dist synced) all green.

## Summary by CodeRabbit

* **New Features**
  * Added clear loading indicators to audit logs, usage data, workflows, guardrails, authentication keys, and overview dashboards.
  * Overview cards, charts, cache meters, and cache statistics now show loading states while data is retrieved.
  * Search fields display an in-progress indicator during filtering.
  * Added English and Polish loading labels.

* **Bug Fixes**
  * UUID searches now match identifier fields accurately without matching unrelated free-text error messages.
  * Improved case-insensitive matching for UUID searches.


SantiagoDePolonia:
Addressed all four review findings in 0c0d120e:

- **Total requests finishes loading early** (greptile P1 + coderabbit): the Total Requests card now waits on `loading || (cacheEnabled && cacheLoading)` since its value folds cache hits in.
- **PG AND-reordering on the error-message gate** (coderabbit): switched to a `CASE WHEN … THEN … ELSE FALSE END` guard — CASE arms are documented not to evaluate unless needed, works identically on SQLite. Re-verified against both engines (live Postgres subtests included).
- **Stale cache response after the gate flips off** (coderabbit): `fetchCacheOverview` now aborts the in-flight controller (and clears `cacheLoading`) in the disabled branch before resetting.
- **Spinner hidden by `role="img"`** (coderabbit): the cache-meter bar only carries `role="img"`/`aria-label` when it's actually showing data, so the loading spinner's `role="status"` reaches the accessibility tree.

PR #708, opened 2026-08-19
feat(openai): add api_mode to translate Responses API via chat completions

zuohuadong:
## Problem

Some OpenAI-compatible upstreams (observed on SGLang) expose `/v1/responses` but return `500 Internal Server Error` for any request with a non-empty `tools` array — even a single minimal function tool. Requests without tools succeed. This makes the Responses API unusable for tool-calling clients (e.g. Codex CLI, which always sends tool definitions) against such upstreams, while the same models serve chat completions with tools correctly.

## Change

Add an opt-in `api_mode` for the generic `openai` provider type. When set to `chat` / `chat_completions` / `chat_compatible` / `responses_via_chat`, `Responses` and `StreamResponses` are routed through the existing `ResponsesViaChat` translation layer (already used by providers embedding `ChatCompatible`) instead of native `/responses` passthrough.

- Default behavior unchanged: empty/absent `api_mode` keeps native passthrough.
- Configurable per provider via YAML `api_mode` or the existing `<PROVIDER>_API_MODE` env var convention.
- Documented in the SGLang provider page.

## Tests

- `TestResponses_APIModeChatCompatibleTranslatesViaChat`: upstream `/responses` returns 500; provider with `api_mode: chat_compatible` succeeds via `/chat/completions` and translates tools into chat-completions function shape.
- `TestResponses_APIModeDefaultKeepsNativePassthrough`: default config still forwards to `/responses`.
- Full `internal/providers/openai` package suite passes.

## Summary by CodeRabbit

* **New Features**
  * Added chat-compatible routing for Responses API requests, enabling function calling when a provider’s native Responses endpoint does not support tool-bearing requests.
  * Native Responses API behavior remains available by default.
* **Documentation**
  * Documented configuration options for enabling chat-compatible mode through YAML or an environment variable.
* **Bug Fixes**
  * Improved compatibility with SGLang versions that reject tool-enabled Responses API requests.

### DIFF
--- a/docs/providers/sglang.mdx
+++ b/docs/providers/sglang.mdx
@@ -194,6 +194,10 @@ GoModel strips client authorization before forwarding and applies
 
 - Chat completions, streaming, model listing, Responses, and embeddings use
   SGLang's OpenAI-compatible API.
+- SGLang's native `/v1/responses` endpoint rejects tool-bearing requests on
+  some versions. Set `api_mode: chat_compatible` in YAML (or
+  `SGLANG_API_MODE=chat_compatible`) to translate Responses API calls through
+  chat completions instead, which keeps function calling working.
 - Embeddings and model-specific features depend on the model loaded by SGLang.
 - Native batch, file, and stored-response lifecycle interfaces are not yet
   exposed as typed GoModel provider capabilities; use passthrough where the
--- a/internal/providers/openai/openai.go
+++ b/internal/providers/openai/openai.go
@@ -2,6 +2,8 @@
 package openai
 
 import (
+	"context"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -27,12 +29,29 @@ const (
 	defaultBaseURL = "https://api.openai.com/v1"
 )
 
+// apiModeResponsesViaChat lists the APIMode values that switch the Responses
+// API from native upstream passthrough to translation over chat completions.
+// OpenAI-compatible upstreams such as SGLang or older vLLM builds expose
+// /v1/responses but reject or fail on tool-bearing requests; routing through
+// chat completions keeps function calling working on those upstreams.
+var apiModeResponsesViaChat = map[string]bool{
+	"chat":               true,
+	"chat_completions":   true,
+	"chat-compatible":    true,
+	"chat_compatible":    true,
+	"responses_via_chat": true,
+	"responses-via-chat": true,
+}
+
 // Provider implements the core.Provider interface for OpenAI.
 // Credentials and the realtime base URL are both read live from the embedded
 // CompatibleProvider, so SetBaseURL overrides and key rotation are honored on
 // the realtime websocket dial target too (see realtime.go).
 type Provider struct {
 	*CompatibleProvider
+	// responsesViaChat translates Responses API calls through chat
+	// completions instead of forwarding to the upstream /responses endpoint.
+	responsesViaChat bool
 }
 
 // New creates a new OpenAI provider.
@@ -44,7 +63,27 @@ func New(cfg providers.ProviderConfig, opts providers.ProviderOptions) core.Prov
 			BaseURL:      baseURL,
 			SetHeaders:   setHeaders,
 		}),
+		responsesViaChat: apiModeResponsesViaChat[strings.ToLower(strings.TrimSpace(cfg.APIMode))],
 	}
+}
+
+// Responses sends a Responses API request. With APIMode set to a
+// chat-compatible value the request is translated through chat completions;
+// otherwise it is forwarded to the upstream /responses endpoint.
+func (p *Provider) Responses(ctx context.Context, req *core.ResponsesRequest) (*core.ResponsesResponse, error) {
+	if p.responsesViaChat {
+		return providers.ResponsesViaChat(ctx, p, req)
+	}
+	return p.CompatibleProvider.Responses(ctx, req)
+}
+
+// StreamResponses streams a Responses API request, honoring the same APIMode
+// translation switch as Responses.
+func (p *Provider) StreamResponses(ctx context.Context, req *core.ResponsesRequest) (io.ReadCloser, error) {
+	if p.responsesViaChat {
+		return providers.StreamResponsesViaChat(ctx, p, req, "openai")
+	}
+	return p.CompatibleProvider.StreamResponses(ctx, req)
 }
 
 // NewWithHTTPClient creates a new OpenAI provider with a custom HTTP client.
--- a/internal/providers/openai/openai_test.go
+++ b/internal/providers/openai/openai_test.go
@@ -37,6 +37,108 @@ func TestNew_ReturnsProvider(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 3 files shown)


SantiagoDePolonia:
Thank you for your PR. This is what my AI agent said after my conversation with it:

> Thanks for pinning this down — the diagnosis is right, the layer is not.
>
> ## Problem
>
> SGLang’s `/v1/chat/completions` supports custom function tools (with `--tool-call-parser`). Its `/v1/responses` does not: custom `type: "function"` tools 500. That is an upstream gap ([sgl-project/sglang#13292](https://github.com/sgl-project/sglang/issues/13292)). Codex talks Responses and always sends function tools, so the dedicated `sglang` provider forwards `/v1/responses` and fails.
>
> Chat-in stays chat-out. Responses-in stays Responses-out at the public API. Only some providers already translate Responses → chat underneath (`ResponsesViaChat`: Groq, Ollama, Gemini, …). SGLang currently does native passthrough.
>
> ## Why this PR’s approach is the wrong surface
>
> Putting `api_mode` on `openai.Provider` does not fix the documented path:
>
> - Default setup is `type: sglang` (`SGLANG_BASE_URL`). `sglang.New` never reads `cfg.APIMode`, so `SGLANG_API_MODE=chat_compatible` is a no-op. Greptile reproduced this after the PR.
> - Official OpenAI’s `/responses` already handles function tools. An opt-in on `type: openai` is unused for real OpenAI and dangerous if flipped: you lose `previous_response_id`, hosted tools, and stored IDs, while the embedded `CompatibleProvider` still advertises native lifecycle (`GetResponse`, …).
> - `api_mode` already means something else (Gemini/Vertex: `native` vs `openai_compatible`; Bedrock Mantle: `auto` / `openai` / `standard`).
>
> `ResponsesViaChat` itself is the right mechanism. It belongs on `sglang`.
>
> Always translating every SGLang Responses call is also too blunt: native `/responses` is useful for no-tools and for SGLang’s hosted/MCP tools (`web_search_preview`, `code_interpreter`). `normalizeResponsesToolsForChat` drops non-function tools.
>
> Mixing `/responses` then `/chat/completions` does **not** corrupt SGLang’s radix/prefix KV cache (token-keyed reuse; a miss is extra prefill, not poison). The real break is `previous_response_id`, which via-chat cannot honor.
>
> ## Proposed implementation
>
> Keep this PR’s idea (opt-in translation) but move it to `internal/providers/sglang` and default it so Codex works without config.
>
> `api_mode` on the **sglang** provider (YAML `api_mode` / `SGLANG_API_MODE`):
>
> | Value | Behaviour |
> |---|---|
> | `auto` (default, empty) | Function tools → `ResponsesViaChat`. Otherwise native `/responses`. |
> | `native` (`responses`) | Always native `/responses`. |
> | `chat_compatible` (`chat`, `responses_via_chat`) | Always via chat. |
>
> `auto` is the good default: function-tool clients work; hosted-tool / no-tool Responses stay native. `native` and `chat_compatible` pin one upstream API when you want a stable prefix for radix cache, or when a newer SGLang build has fixed function tools on `/responses`.
>
> Do **not** change `openai.Provider`. People pointing `type: openai` at SGLang should use `type: sglang`.
>
> Sketch (replace the current `Responses` / `StreamResponses` on `sglang.Provider`):
>
> ```go
> // SGLang /v1/responses accepts hosted/MCP tools but 500s on custom
> // function tools (https://github.com/sgl-project/sglang/issues/13292).
> // Chat completions handle those tools. api_mode selects the bridge:
> //   auto (default): via chat only when the request has function tools
> //   native:         always POST /responses
> //   chat_compatible: always ResponsesViaChat
> const (
> 	responsesModeAuto           = "auto"
> 	responsesModeNative         = "native"
> 	responsesModeChatCompatible = "chat_compatible"
> )
>
> type Provider struct {
> 	compatible    *openai.CompatibleProvider
> 	rootClient    *llmclient.Client
> 	responsesMode string
> }
>
> func New(cfg providers.ProviderConfig, opts providers.ProviderOptions) core.Provider {
> 	// ...existing client setup...
> 	return &Provider{
> 		compatible:    ...,
> 		rootClient:    ...,
> 		responsesMode: normalizeSGLangResponsesMode(cfg.APIMode),
> 	}
> }
>
> func (p *Provider) Responses(ctx context.Context, req *core.ResponsesRequest) (*core.ResponsesResponse, error) {
> 	if p.useChatForResponses(req) {
> 		return providers.ResponsesViaChat(ctx, p, req)
> 	}
> 	return p.compatible.Responses(ctx, req)
> }
>
> func (p *Provider) StreamResponses(ctx context.Context, req *core.ResponsesRequest) (io.ReadCloser, error) {
> 	if p.useChatForResponses(req) {
> 		return providers.StreamResponsesViaChat(ctx, p, req, "sglang")
> 	}
> 	return p.compatible.StreamResponses(ctx, req)
> }
>
> func (p *Provider) useChatForResponses(req *core.ResponsesRequest) bool {
> 	switch p.responsesMode {
> 	case responsesModeNative:
> 		return false
> 	case responsesModeChatCompatible:
> 		return true
> 	default: // auto
> 		return responsesHasFunctionTool(req)
> 	}
> }
>
> func responsesHasFunctionTool(req *core.ResponsesRequest) bool {
> 	if req == nil {
> 		return false
> 	}
> 	for _, tool := range req.Tools {
> 		if t, _ := tool["type"].(string); strings.TrimSpace(t) == "function" {
> 			return true
> 		}
> 	}
> 	return false
> }
>
> func normalizeSGLangResponsesMode(apiMode string) string {
> 	switch strings.ToLower(strings.TrimSpace(apiMode)) {
> 	case "", responsesModeAuto:
> 		return responsesModeAuto
> 	case responsesModeNative, "responses":
> 		return responsesModeNative
> 	case responsesModeChatCompatible, "chat", "chat_completions", "responses_via_chat":
> 		return responsesModeChatCompatible
> 	default:
> 		return responsesModeAuto
> 	}
> }
> ```
>
> Also:
>
> - Advertise `api_mode` on sglang’s credential schema (`auto`, `native`, `chat_compatible`), Advanced.
> - Docs: default is `auto`; show when to pin `native` vs `chat_compatible`. Drop the claim that `SGLANG_API_MODE` works on this PR as written.
> - Tests on `type: sglang`: no tools + `auto` → `/v1/responses`; function tool + `auto` → `/chat/completions`; `native` + tools → `/v1/responses`; `chat_compatible` + no tools → `/chat/completions`; same for `StreamResponses`. Existing no-tools `/v1/responses` test stays green.
>
> ## Residual limit (any via-chat path)
>
> Function tools **plus** `previous_response_id` still cannot work until SGLang implements function tools on `/responses`, or GoModel grows local previous-response expansion. Via-chat should keep returning a clear 400 rather than a 500. Codex turns that send full `input` (the common case) work under `auto`.
>
> Happy to reshape this PR onto `sglang` along these lines.

SantiagoDePolonia:
@zuohuadong Do you plan to work on this fix further?

weselben:
@zuohuadong would be interesting - cause i actually am in need of this / a relating feature :)

PR #718, opened 2026-08-20
feat(chatgpt): route Codex traffic through a ChatGPT subscription

SantiagoDePolonia:
Adds a `chatgpt` provider so Codex can run through GoModel while staying billed to a ChatGPT plan instead of an OpenAI Platform key.

`Codex -> GoModel -> ChatGPT subscription`

## User-visible impact

One env var enables it — the provider is auto-discovered, no YAML needed:

```bash
CHATGPT_API_KEY=$(jq -r .tokens.access_token ~/.codex/auth.json)
```

Serves `/v1/responses` only (chat completions and embeddings return a clear error). Default models `gpt-5.6-sol, gpt-5.5, gpt-5.4`, overridable with `CHATGPT_MODELS`.

## Provider-specific behavior

The Codex backend accepts a narrow Responses dialect, verified against the live endpoint:

- `stream` must be `true`, `store` must be `false`
- `input` must be a list, not a string
- a strict parameter allowlist rejects `temperature`, `top_p`, `max_output_tokens`, `previous_response_id`, `truncation`, `metadata`, `user`, `service_tier`, `top_logprobs`, and any unknown field

Per Postel's law the provider adapts rather than fails: the body is built from an explicit allowlist struct, string inputs are wrapped in the required message list, and a non-streaming `/v1/responses` call is served by streaming upstream and returning the final response object. The gateway's OpenAI-compatible surface is unchanged.

The ChatGPT account ID is derived from the token's own JWT claim, so no second credential is needed.

## Validation

- `codex exec` returned `ok` through `Codex CLI -> GoModel -> Codex backend`, with Codex's real 11-field payload passing the backend's allowlist
- streaming and non-streaming `/v1/responses` both verified
- against the live `chatgpt.com` backend, requests carrying `temperature`/`max_output_tokens`/string input returned `429` (quota) rather than `400`, confirming the adaptation upstream; the account's subscription quota was exhausted during testing, so a live `200` still needs a re-run after the reset

## Docs

New `providers/chatgpt` page, rewritten Codex guide with subscription vs. bring-your-own-provider tabs, plus overview table, README, `.env.template`, and `config.example.yaml`.

## Summary by CodeRabbit

* **New Features**
  * Added ChatGPT subscription access through the Codex backend.
  * Supports authentication, model selection, streaming responses, and account-aware requests.
  * Added environment-variable and YAML configuration examples.
  * Supports the Responses API; unsupported operations are clearly reported.

* **Documentation**
  * Added provider and Codex setup guides covering tokens, models, quotas, limitations, and billing.
  * Updated provider listings to include ChatGPT subscription access.

* **Tests**
  * Added coverage for authentication, streaming, model listing, errors, and unsupported operations.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -427,6 +427,15 @@
 # OPENAI_SESSION_STICKY_KEYS=false
 # OPENAI_BASE_URL=https://api.openai.com/v1
 
+# ChatGPT subscription (Codex backend, /v1/responses only)
+# Billed against your ChatGPT plan instead of OpenAI Platform credit. The key is
+# the access token from `codex login`; it expires roughly every 10 days.
+#   CHATGPT_API_KEY=$(jq -r .tokens.access_token ~/.codex/auth.json)
+# CHATGPT_API_KEY=
+# CHATGPT_BASE_URL=https://chatgpt.com/backend-api/codex
+# Optional model override; defaults to the models a ChatGPT plan can call.
+# CHATGPT_MODELS=gpt-5.6-sol,gpt-5.5,gpt-5.4
+
 # Anthropic
 # Accepts a Console API key (sk-ant-api...) or a Claude subscription OAuth
 # token from `claude setup-token` (sk-ant-oat...; Claude Code traffic only).
--- a/.env.template
+++ b/.env.template
@@ -434,7 +434,7 @@
 # CHATGPT_API_KEY=
 # CHATGPT_BASE_URL=https://chatgpt.com/backend-api/codex
 # Optional model override; defaults to the models a ChatGPT plan can call.
-# CHATGPT_MODELS=gpt-5.6-sol,gpt-5.5,gpt-5.4
+# CHATGPT_MODELS=gpt-5.6-sol,gpt-5.6-terra,gpt-5.6-luna,gpt-5.5
 
 # Anthropic
 # Accepts a Console API key (sk-ant-api...) or a Claude subscription OAuth
--- a/README.md
+++ b/README.md
@@ -158,8 +158,9 @@ GoModel supports OpenAI, Anthropic, Cohere, Google Gemini, Vertex AI, DeepSeek,
 Groq, Fireworks AI, Meta (Muse Spark), OpenRouter, Z.ai, xAI (Grok), Alibaba
 Cloud Model Studio (Bailian), Kilo AI, MiniMax, Xiaomi MiMo, OpenCode Go, Azure
 OpenAI, Oracle, Ollama, SGLang, vLLM, llm-d, Amazon Bedrock Runtime, Amazon
-Bedrock Mantle, and all OpenAI-compatible providers. Voice: ElevenLabs
-(text-to-speech and speech-to-text).
+Bedrock Mantle, and all OpenAI-compatible providers. Subscription-billed:
+ChatGPT (the Codex backend) and Claude. Voice: ElevenLabs (text-to-speech and
+speech-to-text).
 
 See the [Providers Overview](https://gomodel.enterpilot.io/docs/providers/overview?utm_source=readme) for the full
 per-provider feature matrix (chat, `/responses`, embeddings, files, batches,
--- a/config/config.example.yaml
+++ b/config/config.example.yaml
@@ -337,6 +337,14 @@ providers:
     type: anthropic
     api_key: "sk-ant-..."
 
+  # ChatGPT subscription (Codex backend). Serves /v1/responses only and is
+  # billed against your ChatGPT plan. The key is the access token from
+  # `codex login`: jq -r .tokens.access_token ~/.codex/auth.json
+  chatgpt:
+    type: chatgpt
+    api_key: "${CHATGPT_API_KEY}"
+    # models: [gpt-5.6-sol, gpt-5.5, gpt-5.4]
+
   cohere:
     type: cohere
     api_key: "${COHERE_API_KEY}"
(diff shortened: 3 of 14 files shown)


SantiagoDePolonia:
Reviewed all three findings. Two fixed, one declined.

**Fixed — incomplete streams reported as success** (CodeRabbit `stream.go:55`, partially Greptile `stream.go:49`)

`collapseResponsesStream` kept the last envelope it saw, so a stream that ended early — a dropped connection or an `error` event — returned the `response.created` envelope with `status: in_progress` and empty output, as a successful HTTP 200. Real bug. Only the terminal lifecycle events now produce a response; premature EOF and `error` events return a provider error carrying the upstream message. Added `response.incomplete` as terminal too, which the original missed. Three regression cases added.

**Declined — returning an error for `response.failed`** (Greptile)

A terminal `response.failed` stays a response. That is what the Responses API itself returns for a non-streaming call: HTTP 200 with `status: "failed"` and a populated `error` object, and it is what `CompatibleProvider.Responses` already does for every other provider in this repo. Turning it into a transport error here would make `chatgpt` behave differently from `openai` for the same upstream condition. Covered by a test that pins the behavior.

**Fixed — `gpt-5.4` retirement** (CodeRabbit `chatgpt.go:31`)

Confirmed against [OpenAI's announcement](https://x.com/OpenAIDevs/status/2083301245390143831): `gpt-5.4` and `gpt-5.4-mini` leave ChatGPT-authenticated Codex on 2026-08-31, while staying on the API and API-key-authenticated Codex. Default inventory is now `gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna, gpt-5.5` — the two replacements were verified accepted by the live backend. Updated in the provider, docs, `.env.template`, and `config.example.yaml`, with a note pointing API-key users at the `openai` provider.

**Not actioned — docstring coverage warning**

The 60% figure counts test functions with self-describing names. Every exported declaration has a doc comment; the only undocumented non-test items are the interface assertion and a three-line `unsupported` helper. Consistent with the surrounding provider packages.

Also fixed while re-testing: the guide's inline `codex exec` example omitted the provider `name` field, which Codex rejects. Re-ran the full chain after the changes — `codex exec` still returns `ok` through `Codex -> GoModel -> Codex backend`.

SantiagoDePolonia:
Both minor findings on `dbc873f6` fixed in `39920637`.

**`response.incomplete` coverage** — valid gap. It was added as a terminal event but only `response.failed` was tested. The terminal-response test is now a table over both, asserting each returns without a transport error and preserves its status, plus a case pinning that `failed` carries the upstream error object.

While adding it I renamed `TestResponses_IncompleteStreamIsAnError` to `TestResponses_TruncatedStreamIsAnError`. The two now sit next to each other and mean opposite things: a stream that stops early is an error, whereas a `response.incomplete` event is a legitimate response.

**Tense** — correct, the date is 11 days out. Reworded in `chatgpt.go` and `docs/providers/chatgpt.mdx`.

Full suite, race tests, and lint pass.

SantiagoDePolonia:
Two comments on `88406be7`.

**CodeRabbit — assert the provider error code.** Valid, fixed in `3bf40f3b`. The test pinned the 501 status but not the code, leaving the programmatic half of the contract free to drift. Also extracted the literal into a named `unsupportedOperationCode` constant so the provider and its test reference one source rather than two copies of a string.

**Greptile `stream.go:52`** — no action. The comment reports its own tooling failure ("T-Rex ran the requested verification, but its local artifact references were not uploaded... it did not return a separate root-cause sentence") rather than a finding. There is no claim to verify. I re-read the terminal-event handling anyway and it is correct: the three terminal lifecycle events return the response, `error` events and premature EOF return a provider error. That path has regression coverage from `dbc873f6`/`39920637`. Happy to look again if the check re-runs and produces an actual finding.

Full suite, race tests, and lint pass.

SantiagoDePolonia:
Two new comments.

**CodeRabbit — the override example covered only one of four models.** Real hazard, and worse than described. I checked: with one of the four declared, `GET /v1/models` returns **one** model, not four with stale metadata — the declaration replaces the inventory rather than filtering it. (For this provider that is its own `CHATGPT_MODELS` behaviour, not the generic `fallback`/`allowlist` mode.) The YAML example itself was already gone as of `fffa0aca`, which replaced it with a link to `/advanced/model-metadata`, so the fix is a sentence rather than four more YAML entries: `436f9fef` says to declare every model you want served, because the list replaces the default inventory.

**Greptile `stream.go:52` — declining again, now with evidence.** This is the same finding from `3a51a4cb`, and the comment itself notes the behaviour is intentional and test-pinned. My earlier answer argued from reading `CompatibleProvider.Responses`; since T-Rex ran code, here is the matching experiment.

I pointed the **`openai`** provider at an upstream returning a non-streaming `200` with `status: "failed"` — what the Responses API actually returns for a failed non-streaming generation:

```
gateway HTTP status: 200
body status : failed
body error  : boom
```

Identical to `chatgpt`. Returning an error here would make `chatgpt` the only provider in the gateway that converts a failed generation into a transport error, so the same upstream condition would produce different client-visible results depending on which provider served it. That inconsistency is worse than the 200 it would replace. A failed generation that consumed tokens is also genuinely billable, so recording its usage is not obviously wrong either.

If the concern is that operators cannot distinguish failed generations in usage data, that is worth solving — but gateway-wide, on the shape every provider shares, not by making one provider diverge.

PR #727, opened 2026-08-21
fix(bedrock): surface prompt-cache usage on Anthropic-format responses

### LINKED ISSUE #726: Bedrock: prompt-cache usage missing from Anthropic-format responses (streaming drops it entirely; non-streaming loses cache-write)
Using GoModel as an Anthropic-compatible gateway in front of AWS Bedrock (`/v1/messages`, Claude models via cross-region inference profiles), prompt caching itself works — Bedrock reports cache hits — but the usage fields don't reach Anthropic-format clients in two cases:

**1. Streaming responses carry no cache usage at all.**
`streamConverter.formatChunk` (`internal/providers/bedrock/chat_stream.go`) forwards only `prompt_tokens` / `completion_tokens` / `total_tokens` from the Converse metadata `TokenUsage`, dropping `CacheReadInputTokens` / `CacheWriteInputTokens`. The anthropicapi stream converter is already prepared to surface them (`chatUsage` has `cache_read_input_tokens` / `cache_creation_input_tokens` json tags) — the keys just never arrive. Agentic SDK clients (Claude Agent SDK / Claude Code) always stream, so they never see cache usage from Bedrock.

**2. Non-streaming responses lose cache-write usage to a key-name mismatch.**
`bedrockUsageExtras` (`internal/providers/bedrock/chat.go`) emits `cache_write_input_tokens`, but `internal/anthropicapi/response.go` reads `cache_creation_input_tokens` (the key the `anthropic` provider emits). Nothing reads `cache_write_input_tokens`, so `cache_creation_input_tokens` is always absent from `/v1/messages` responses.

**Repro** (Bedrock provider, any Claude model, a system block above the model's minimum cacheable size with `cache_control: {"type": "ephemeral"}`):
- Non-streaming, first call: usage lacks `cache_creation_input_tokens` (expected ≈ block size).
- Second call: `cache_read_input_tokens` IS present — confirming the cache works and only reporting is affected.
- Same two calls with `"stream": true`: `message_delta` usage carries neither field.

**Impact:** cost accounting built on `/v1/messages` usage misstates real spend for cached Bedrock workloads, and cache hit-rate monitoring reads permanently zero for streaming clients.

The fix is small — PR incoming.

roshaans:
Fixes #726.

Two small changes so Bedrock prompt-cache counters reach Anthropic-format (`/v1/messages`) clients:

- **`bedrockUsageExtras`**: also emit `cache_creation_input_tokens` — the canonical key `internal/anthropicapi/response.go` reads and the `anthropic` provider emits — alongside the existing `cache_write_input_tokens` (kept for anything that learned the old key).
- **`streamConverter.formatChunk`**: forward `CacheReadInputTokens` / `CacheWriteInputTokens` from the Converse metadata usage into the chunk usage payload under the canonical keys, so streamed responses carry cache usage like non-streaming ones. `anthropicapi`'s `chatUsage` already parses both keys — no changes needed there.

**Tests**: three new unit tests — `formatChunk` forwards cache fields when Bedrock reports them, omits them when absent, and `bedrockUsageExtras` emits the canonical + legacy write keys. `go test ./internal/providers/bedrock/... ./internal/anthropicapi/...` passes; `go build ./...` and `go vet` clean.

**Verified against live Bedrock** (Claude Haiku 4.5 via a `us.` cross-region inference profile, cached system block): streamed `/v1/messages` responses now report `cache_read_input_tokens` (5287 in our test) where they previously carried no cache usage; non-streaming first calls now report `cache_creation_input_tokens`.

## Summary by CodeRabbit

* **New Features**
  * Added prompt-cache usage details to streamed responses.
  * Cache read and cache creation token counts are now reported using standardized usage fields.
  * Retained the existing cache-write usage field for compatibility.

### DIFF
--- a/internal/providers/bedrock/bedrock_test.go
+++ b/internal/providers/bedrock/bedrock_test.go
@@ -893,3 +893,68 @@ func TestCachePointFallbackIsNarrowAndLossless(t *testing.T) {
 		t.Fatal("cache-point removal mutated original parts")
 	}
 }
+
+func TestStreamConverter_FormatChunkForwardsCacheUsage(t *testing.T) {
+	sc := newOpenAIStream(nil, "test-model")
+	chunk := sc.formatChunk(map[string]any{}, "stop", &brtypes.TokenUsage{
+		InputTokens:           awssdk.Int32(3),
+		OutputTokens:          awssdk.Int32(7),
+		TotalTokens:           awssdk.Int32(10),
+		CacheReadInputTokens:  awssdk.Int32(5000),
+		CacheWriteInputTokens: awssdk.Int32(1200),
+	})
+	payload := strings.TrimSuffix(strings.TrimPrefix(chunk, "data: "), "\n\n")
+	var parsed map[string]any
+	if err := json.Unmarshal([]byte(payload), &parsed); err != nil {
+		t.Fatalf("payload not JSON: %v", err)
+	}
+	usage, ok := parsed["usage"].(map[string]any)
+	if !ok {
+		t.Fatalf("usage missing: %v", parsed)
+	}
+	if usage["cache_read_input_tokens"].(float64) != 5000 {
+		t.Errorf("cache_read_input_tokens = %v, want 5000", usage["cache_read_input_tokens"])
+	}
+	if usage["cache_creation_input_tokens"].(float64) != 1200 {
+		t.Errorf("cache_creation_input_tokens = %v, want 1200", usage["cache_creation_input_tokens"])
+	}
+}
+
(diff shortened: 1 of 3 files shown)


roshaans:
Addressed in e07a31e0: reworked the cache-usage tests into table-driven cases covering both-present, read-only, write-only, and both-absent for `formatChunk` and `bedrockUsageExtras`, asserting present fields are emitted and absent ones omitted.

SantiagoDePolonia (approved):
Looks good to me!

SantiagoDePolonia:
@roshaans Thank you for reporting and fixing! It'll be released later today

PR #751, opened 2026-08-24
fix(vllm): rename legacy reasoning_content to reasoning on replayed assistant messages

MacherelR:
## Summary

- vLLM >=0.20 (vllm-project/vllm RFC #27755) renamed the reasoning field
  from `reasoning_content` to `reasoning` on incoming assistant messages,
  and silently discards the legacy key — no error, it just never reaches
  the chat template.
- Clients built against the older field name replay real prior-turn
  reasoning under `reasoning_content`, intending to preserve it for
  Qwen3-family "preserve thinking" behavior, but that content is dropped
  on the floor before vLLM ever sees it.
- Add an `AdaptChatRequest` hook for the vLLM provider (same extension
  point the DeepSeek provider uses) that renames `reasoning_content` ->
  `reasoning` on replayed assistant messages, only when `reasoning` isn't
  already present. No-op for clients already on the current field name or
  sending neither field.

## Context

Found while debugging why Qwen3.8-27B (served via vLLM 0.20.0, behind
GoModel) wasn't benefiting from multi-turn reasoning continuity. Empirically
confirmed via GoModel's own `audit_logs` that:

- LibreChat and Crush never send any reasoning field on replayed assistant
  messages at all.
- OpenCode sends real reasoning content, but under the legacy
  `reasoning_content` key — silently dropped by vLLM 0.20+.

This PR fixes the OpenCode case (real content, wrong key). The LibreChat/
Crush case is a different problem — no reasoning sent by the client at all —
and would need GoModel to reconstruct history server-side from its own
stored response bodies. Out of scope here; happy to follow up separately if
there's interest.

## Testing

- `go build ./...`
- `go vet ./internal/providers/vllm/...`
- `golangci-lint run ./internal/providers/vllm/...`
- `go test ./internal/providers/vllm/...` — all passing, including 5 new
  tests: rename-on-replay, don't-override-existing-`reasoning`, no-op
  without the legacy field, non-`assistant` messages ignored, nil request.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vLLM compatibility by automatically translating legacy assistant reasoning fields into the supported format.
  - Preserves existing reasoning data and leaves unrelated messages and requests unchanged.
  - Applies consistently across standard and custom HTTP client configurations.

- **Tests**
  - Added coverage for field translation, preservation, no-op scenarios, non-assistant messages, nil requests, request immutability, and outgoing payloads.

### DIFF
--- a/internal/providers/vllm/reasoning.go
+++ b/internal/providers/vllm/reasoning.go
@@ -1,0 +1,59 @@
+// Package vllm provides vLLM OpenAI-compatible API integration for the LLM gateway.
+package vllm
+
+import (
+	"github.com/goccy/go-json"
+
+	"github.com/enterpilot/gomodel/internal/core"
+)
+
+// adaptChatRequest renames the legacy reasoning_content field to reasoning on
+// replayed assistant messages.
+//
+// vLLM >=0.20 (vllm-project/vllm RFC #27755) reads only the "reasoning"
+// field on incoming assistant messages and silently discards
+// "reasoning_content" before its chat template ever runs. Clients built
+// against the older field name (e.g. OpenCode) intend to replay a prior
+// turn's thinking so Qwen3-family "preserve thinking" behavior keeps
+// working across turns, but that content is dropped on the floor with no
+// error, degrading multi-turn/agentic quality silently.
+//
+// Renaming here restores that continuity without requiring a client change.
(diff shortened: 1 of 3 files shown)


SantiagoDePolonia:
@MacherelR Thank you for fixing this!

It'll be released today.

Don't hesitate to report any issues or to propose any feature!

PR #769, opened 2026-08-25
feat(models): filter provider inventory by glob pattern and price cap

### LINKED ISSUE #757: Feature: OpenRouter (or in general?) model globs, or price selection
**Is your feature request related to a problem? Please describe.**

I use OpenRouter to experiment with free models and OpenAI and local models for actually doing something useful. OpenRouter has lots of models, but the free ones have :free suffix.

**Describe the solution you'd like**

I would like to have either cost control for OpenRouter model selection (e.g. do not use models more expensive than X), or just ability to choose :free models only for the provider I use for experiments.

SantiagoDePolonia:
Closes #757.

Adds `providers.<name>.model_filter` so a provider's catalog can be narrowed to the models that should be routable, instead of listing every model by hand. The motivating case from the issue: run OpenRouter's free tier for experiments while other providers serve real work.

```yaml
providers:
  openrouter:
    type: openrouter
    api_key: "${OPENROUTER_API_KEY}"
    model_filter:
      include: ["*:free"]
      exclude: ["*-preview:free"]
      max_price_per_mtok: 0
```

| Field | Default | Behavior |
| --- | --- | --- |
| `include` | empty (keep all) | Keeps only models matching at least one pattern. |
| `exclude` | empty | Drops models matching any pattern. Applied after `include`. |
| `max_price_per_mtok` | unset (no cap) | Drops models whose highest per-million-token rate — the larger of input and output — is above the cap. |

Same rules in env, per provider: `OPENROUTER_MODEL_FILTER_INCLUDE`, `..._EXCLUDE`, `..._MAX_PRICE_PER_MTOK`. Each rule overlays YAML independently, so an env price cap narrows a YAML pattern filter without restating it.

## User-visible impact

- No behavior change without configuration: an unset filter keeps every model.
- Filtered models leave the catalog entirely — absent from `GET /v1/models`, and requests for them return "model not found". That is the point of the feature, but it is a routing change, not just a listing change.
- **A model with no known pricing is dropped while `max_price_per_mtok` is set.** A cap that admits models of unknown price is not a cap, so this fails closed. Local providers usually report no pricing at all — filter those by pattern, not by price.
- Patterns are globs matched case-insensitively against the raw provider model ID. `*` matches any run of characters **including `/`** (unlike `filepath.Match`), so `*:free` matches `deepseek/deepseek-r1:free`.

## Provider-specific behavior

OpenRouter's `/models` payload carries per-token prices that `ListModels` was discarding. It now maps them to per-Mtok metadata, so the `:free` tier is priced at zero without any registry entry. Rates of `-1` (auto-routed models, where OpenRouter cannot state a price up front) are reported as unpriced rather than as a negative price that would look cheaper than free. Enrichment already treats what a provider reports as the override, so this also fixes `strategy: cost` load balancing on OpenRouter, which previously fell back to "first target" for most of that catalog.

## Implementation notes

- Filtering runs after metadata enrichment, so price rules see resolved pricing, and after `configured_provider_models_mode` resolves, so it also narrows models added by `models` in merge/allowlist mode.
- It re-runs on live re-enrichment (`EnrichModels`): a model that becomes too expensive when the remote model list refreshes leaves the catalog then, rather than staying routable until the next provider fetch sweep.
- A filter that matches nothing leaves the provider present with an empty inventory. Dropping the provider key would read as a failed refresh and resurrect the previous inventory through the carry-forward path; the case is logged at WARN instead.

## Not included

Glob patterns as *virtual model targets* (a `free` alias that round-robins across every `:free` model) are the natural companion and reuse the same matcher, but they touch the virtual-model snapshot and catalog interface and are kept out of this change.

## Summary by CodeRabbit

* **New Features**
  * Added provider model filtering with case-insensitive include/exclude glob patterns.
  * Added optional maximum price limits per million tokens.
  * Filters apply to fetched, configured, and cached model inventories.
  * Models without pricing are excluded when a price limit is active.
  * OpenRouter pricing metadata now supports filtering, including free models.

* **Documentation**
  * Added YAML and environment-variable configuration guidance, examples, validation rules, and provider references.
  * Documented filtering behavior and interactions with model inventory settings.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -216,7 +216,8 @@
 # INCLUDE keeps only matching models; EXCLUDE then drops matching models.
 # MAX_PRICE_PER_MTOK caps a model's highest per-million-token rate (the larger of
 # input and output). A model with no known price is dropped while the cap is set.
-# The cap must be a finite number of 0 or more; anything else fails startup.
+# The cap must be a finite number of 0 or more; anything else (including a value
+# that is not a number at all) fails startup rather than leaving the cap unset.
 # Set per provider: <PROVIDER>[_SUFFIX]_MODEL_FILTER_INCLUDE, ..._EXCLUDE, ..._MAX_PRICE_PER_MTOK
 # OPENROUTER_MODEL_FILTER_INCLUDE=*:free
 # OPENROUTER_MODEL_FILTER_EXCLUDE=*-preview:free
--- a/.env.template
+++ b/.env.template
@@ -216,6 +216,7 @@
 # INCLUDE keeps only matching models; EXCLUDE then drops matching models.
 # MAX_PRICE_PER_MTOK caps a model's highest per-million-token rate (the larger of
 # input and output). A model with no known price is dropped while the cap is set.
+# The cap must be a finite number of 0 or more; anything else fails startup.
 # Set per provider: <PROVIDER>[_SUFFIX]_MODEL_FILTER_INCLUDE, ..._EXCLUDE, ..._MAX_PRICE_PER_MTOK
 # OPENROUTER_MODEL_FILTER_INCLUDE=*:free
 # OPENROUTER_MODEL_FILTER_EXCLUDE=*-preview:free
--- a/.env.template
+++ b/.env.template
@@ -208,6 +208,19 @@
 # CONFIGURED_PROVIDER_MODELS_MODE=fallback
 # Examples: OPENROUTER_MODELS=..., OPENROUTER_EU_MODELS=..., AZURE_MODELS=..., VLLM_MODELS=...
 
+# Narrow a provider's model inventory to what you actually want routable. Applied
+# to the final inventory, so it also narrows models added by <PROVIDER>_MODELS.
+# Patterns are comma-separated globs matched case-insensitively against the raw
+# model ID: `*` matches any run of characters (including `/`, so `*:free` matches
+# `deepseek/deepseek-r1:free`) and `?` matches exactly one.
+# INCLUDE keeps only matching models; EXCLUDE then drops matching models.
+# MAX_PRICE_PER_MTOK caps a model's highest per-million-token rate (the larger of
+# input and output). A model with no known price is dropped while the cap is set.
+# Set per provider: <PROVIDER>[_SUFFIX]_MODEL_FILTER_INCLUDE, ..._EXCLUDE, ..._MAX_PRICE_PER_MTOK
+# OPENROUTER_MODEL_FILTER_INCLUDE=*:free
+# OPENROUTER_MODEL_FILTER_EXCLUDE=*-preview:free
+# OPENROUTER_MODEL_FILTER_MAX_PRICE_PER_MTOK=0
+
 # Virtual models as infrastructure-as-code (JSON array). Declares redirects, load
 # balancers, and access policies; overrides admin-store rows with the same source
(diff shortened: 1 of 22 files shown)


SantiagoDePolonia:
Addressed all five findings in 7afa8f33.

**Filtering permanently removed models (Greptile P1)** — valid, and the more serious half was the cache: `SaveToCache` persisted the filtered inventory, so a loosened filter was unrecoverable across a restart with the provider unreachable.

Filtering is now a view rather than a deletion. The registry keeps `discoveredByProvider` (unfiltered, what enrichment runs over and what is persisted) and derives `modelsByProvider` from it through the filters. Republishing happens after enrichment, on every inventory apply, and when a filter is set or cleared. A provider with no filter shares its inner map, so the no-filter path costs nothing beyond the outer map. Regression tests cover re-admission after a price drop, clearing a filter, and the cache round-trip with the filter removed.

**Non-finite / negative price caps (both reviewers)** — valid. `NaN` rejected everything, `+Inf` disabled the cap, negative could never be met. These now fail startup with a message naming the provider, validated after the env overlay so one rule covers YAML and `<PROVIDER>_MODEL_FILTER_MAX_PRICE_PER_MTOK` alike. `perMtok` in the OpenRouter listing rejects `NaN`, `±Inf`, and per-Mtok scaling overflow, so a corrupt rate reports no price rather than a wrong one.

An unparseable env value (`=cheap`) is still ignored, matching the existing convention for `SESSION_STICKY_KEYS` and `FAIRNESS_FROM_USER_PATH` in the same parser.

**`total_models` double-counting (CodeRabbit)** — valid. Both paths now count what the catalog publishes instead of subtracting per-provider removals: `len(r.models)` on initialize, and distinct published IDs across the refreshed providers on a targeted refresh, which keeps that log field scoped to the refresh as before.

**Test env hygiene (CodeRabbit)** — fixed, the three `OPENROUTER_MODEL_FILTER_*` variables are in `clearProviderEnvVars`.

`make lint` clean, full `go test ./...` and `-race` on the touched packages green.

SantiagoDePolonia:
Both addressed in 40fd20e5.

**Malformed env price cap (Greptile P1)** — you're right, and this is the one I'd explicitly left in the previous round. My reasoning was consistency with `SESSION_STICKY_KEYS` and `FAIRNESS_FROM_USER_PATH`, which ignore unparseable input in the same parser. That argument doesn't hold here: those are booleans that fall back to a safe default, whereas a price cap that quietly disappears is a cost control the operator believes is in force. Failing closed wins over local convention.

`<PROVIDER>_MODEL_FILTER_MAX_PRICE_PER_MTOK=cheap` now carries NaN into `ModelFilter.Validate`, which rejects it at startup with the same message as a literal `NaN` — the value is not a number, so it is reported as one. Covered at both levels: the env parser, and `resolveProviders` → `validateProviderModelFilters` end to end, so the value cannot be dropped somewhere between parsing and validation.

A malformed cap on a provider with no resolvable credentials is still dropped by `filterEmptyProviders` before validation, as before — that provider is never registered, so nothing routes through it.

**Model-filter variables cleared for one provider only (CodeRabbit)** — valid, `parseProviderEnvKey` accepts the three fields for every prefix. Rather than adding 3 entries per provider by hand, `clearProviderEnvVars` now derives the prefixes from its own list and generates the filter variables for each, so a provider added later is covered automatically.

`make lint` clean, full `go test ./...` green.

PR #771, opened 2026-08-25
fix(cache): fall back to local model cache when Redis is unreachable

### LINKED ISSUE #630: Issue: GoModel should degrade gracefully when Redis is unavailable
**Describe the bug**
If `cache.model.redis.url` (or `REDIS_URL`) is set but Redis is unreachable at startup, GoModel fails to start entirely instead of falling back to the local file cache or starting in a degraded mode. This makes Redis a hard single point of failure for the whole gateway, even though a local cache backend already exists and works fine on its own.

Root cause: `initCache` (internal/providers/init.go, around lines 187-219) only tries the Redis branch when `m.Redis.URL != ""` and returns the connection error straight up (`return nil, err` at line 201) with no fallback to `m.Local`, even when a local cache config is also present. This error propagates up through provider initialization and the whole application fails to start (`failed to initialize application: failed to initialize providers: failed to initialize cache: failed to connect to redis: ...`).

This also affects the response cache path (`cache.response.simple`) - see internal/responsecache/responsecache.go, which currently only warns (not hard-fails) when the URL is missing entirely, but its behavior on a configured-but-unreachable Redis at runtime (after startup, mid-operation) has not been separately audited here and may need its own look once the startup path is fixed.

**GoModel version**
v0.1.68 (HEAD around 9d54827f, 2026-08-03)

**Steps To Reproduce**
1. Set `REDIS_URL` to a Redis instance that is not running/reachable.
2. Start GoModel.
3. Observe: the process logs `redis: connection pool: failed to dial after 5 attempts` several times, then `ERROR failed to initialize application error="failed to initialize providers: failed to initialize cache: failed to connect to redis: ..."` and exits/fails to serve any traffic at all.

**Screenshots/Logs**
```
{"level":"WARN","msg":"redis: connection pool: failed to dial after 5 attempts: dial tcp [::1]:6379: connect: connection refused"}
{"level":"ERROR","msg":"failed to initialize application","error":"failed to initialize providers: failed to initialize cache: failed to connect to redis: dial tcp [::1]:6379: connect: connection refused"}
```

**Additional context**
This is a bigger design task, not a one-line fix:
- Decide the desired behavior: fall back to the local file cache when Redis is configured but unreachable at startup (with a clear warning log), rather than a hard failure.
- Handle the case where Redis becomes unreachable later, at runtime (after a successful startup) - reconnect handling, not just a one-time startup check.
- Apply the same graceful-degradation thinking to the other Redis-backed features (response cache, semantic cache) for consistency, even though this issue is scoped primarily at the model-cache startup failure described above.
- Needs tests covering: Redis unreachable at startup with local cache configured (should start successfully in degraded mode), Redis unreachable with no local cache configured (should still fail clearly, since there is genuinely no cache backend available), and Redis recovering after being down.

BetterAndBetterII:
If `cache.model.redis` is set but Redis is unreachable at startup, `initCache` used to return the connection error and abort process init even when a local file cache was also configured.

Keep Redis as the preferred backend. When the Redis client cannot connect and a local backend is present, log a warning and continue on the local file cache. Redis-only configurations still fail so a missing cache is not silent.

Config validation now allows both backends together so a YAML local cache can stay in place as the fallback when `REDIS_URL` is also set.

Fixes #630

## Summary by CodeRabbit

* **New Features**
  * Redis is now preferred for model caching when configured.
  * Local caching automatically provides a fallback if Redis is unavailable.
  * Configurations may include both Redis and local cache settings.

* **Bug Fixes**
  * Applications no longer fail to start when Redis is unreachable, provided a local cache is configured.
  * Configurations without any cache backend remain invalid.

* **Documentation**
  * Configuration guidance now explains Redis-first behavior and local-cache fallback.

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -15,7 +15,8 @@ type CacheConfig struct {
 }
 
 // ModelCacheConfig holds cache configuration for model registry.
-// Exactly one of Local or Redis must be non-nil.
+// At least one of Local or Redis must be non-nil. When both are set, Redis is
+// preferred and Local is the fallback if Redis is unreachable.
 type ModelCacheConfig struct {
 	RefreshInterval int `yaml:"refresh_interval" env:"CACHE_REFRESH_INTERVAL"`
 	// RecheckInterval is how often (seconds) providers whose latest refresh
@@ -135,10 +136,11 @@ type WeaviateConfig struct {
 }
 
 // ValidateCacheConfig validates the cache configuration in c.
-// For the model cache, exactly one backend (Local or Redis) must be configured;
-// having both or neither is an error. When Redis is selected, its URL must be
-// non-empty. Returns a descriptive error if any constraint is violated, or nil
-// if the configuration is valid.
+// For the model cache, at least one backend (Local or Redis) must be configured;
+// having neither is an error. Both may be set: Redis is preferred and Local is
+// the fallback if Redis is unreachable at startup. When Redis is selected, its
+// URL must be non-empty. Returns a descriptive error if any constraint is
+// violated, or nil if the configuration is valid.
 func ValidateCacheConfig(c *CacheConfig) error {
 	if c == nil {
 		return fmt.Errorf("cache: configuration is required")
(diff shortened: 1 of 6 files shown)


SantiagoDePolonia:
@BetterAndBetterII Thank. you for your contribution!

Merged!

PR #784, opened 2026-08-27
fix(release): publish core.txt with a deploy key instead of a dispatch

SantiagoDePolonia:
`WEBSITE_DISPATCH_TOKEN` was never set, so the dispatch step failed on every release — v0.1.83 included:

```
env:
  GH_TOKEN:
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable
##[error]Process completed with exit code 4
```

Nobody noticed because the website's nightly cron regenerates `core.txt` from the public releases API regardless, so the manifest was only ever up to a day stale — never wrong.

A `repository_dispatch` needs an API token, and a PAT cannot be minted without a human. A **deploy key** can, and it is narrower: one repository, git transport only, no user identity, revocable on its own. So this pushes the manifest instead of dispatching — and the push is itself a deploy trigger, so the site still rebuilds immediately and its sync script derives the same value from the releases API.

`WEBSITE_PUSH_KEY` is already set in this repository, with the matching write deploy key registered on the website repo.

The step mirrors the Pro manifest publish, guards included: refuses to walk the advertised version backwards (older tag re-run, or two releases landing out of order), stages before comparing so an untracked manifest is not read as unchanged, retries against a freshly fetched tip, and serializes on a fixed concurrency group.

Tested by extracting the step body from this YAML and running it against a real git remote: first publish, same-version re-run, newer release, older-tag re-run, and a release published over its own `-rc1` — all correct.

https://claude.ai/code/session_01H8KkD5FSAftZa9dZYrtbQu

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed release manifest publishing so version updates are reliably delivered to the website.
  * Prevented older releases from overwriting newer published versions.
  * Added retry handling to improve reliability during concurrent releases.

* **Chores**
  * Release publishing now serializes concurrent updates to avoid conflicts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -207,24 +207,111 @@ jobs:
   publish-version-manifest:
     name: Publish version manifest
     runs-on: ubuntu-latest
-    # Runs after the GitHub release exists: the website regenerates
-    # public/version/core.txt from the latest release, so dispatching earlier
-    # would republish the previous version.
+    # Runs after the GitHub release exists: the push triggers a website deploy,
+    # whose sync script regenerates core.txt from the latest release. Running
+    # earlier would let that regeneration overwrite this version with the
+    # previous one.
     needs: github-release
-    # Prereleases (v1.2.3-rc1) are not "latest" on GitHub, so the website
-    # would regenerate the same manifest it already serves.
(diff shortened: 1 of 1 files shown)


SantiagoDePolonia:
stale

PR #794, opened 2026-08-29
feat(providers): route provider requests through a per-provider HTTP/SOCKS5 proxy

### LINKED ISSUE #671: Feature: Http/Socks Proxy
It would be nice if we could define/add HTTP/SOCKS5 proxies in admin panel, and when adding each key, we could assign a proxy to a key. So that when we are calling a provider with that certain key, the GoModel would use the proxy information, to call the provider.

SantiagoDePolonia:
Closes #671.

Adds a `proxy_url` setting per provider instance so each provider (and therefore each API key set) can call its upstream through its own forward proxy. Supports `http://`, `https://`, `socks5://` and `socks5h://`, with optional `user:password@` credentials.

Set it any of three ways:
- `config.yaml`: `providers.<name>.proxy_url`
- env: `<PROVIDER>[_SUFFIX]_PROXY_URL` (e.g. `OPENAI_EU_PROXY_URL`)
- dashboard: new **Proxy URL** field under *Advanced settings* on the Providers page, offered for every provider type

## How it works

The factory validates the URL and attaches it to each request's context through the existing `OnRequestStart` hook; the shared `httpclient` transport's `Proxy` callback prefers that over `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`. Every adapter is covered (including `/models` discovery) without changing any of them. Go's transport keys idle connections by proxy, so one transport stays safe.

## User-visible impact

- Default is unchanged: no proxy unless configured.
- Admin API and provider status mask the proxy password (`socks5://user:xxxxx@…`); sending that masked form back on an edit keeps the stored password, like `***` for API keys.
- Storage: new `proxy_url` column/field (SQL migration via `AddColumns`, Mongo field).
- Vertex-backed providers: the Google OAuth token exchange is not routed through the provider proxy — only the API calls are. Documented.

Docs: providers overview (new section), configuration reference, `config.example.yaml`, `.env.template`, OpenAPI.

## Summary by CodeRabbit

* **New Features**
  * Configure an individual provider to use an HTTP, HTTPS, SOCKS5, or SOCKS5H proxy.
  * Set proxy URLs in provider settings, YAML configuration, or environment variables.
  * Manage proxy URLs directly from the provider credentials dashboard.
  * Continue using direct connections or process-wide proxy settings when no provider proxy is configured.
* **Security**
  * Proxy passwords are masked in dashboard views and API responses.
* **Documentation**
  * Added configuration guidance, supported formats, credential handling, precedence rules, and proxy health behavior.


SantiagoDePolonia:
Stale. I would start from scratch

PR #800, opened 2026-08-29
fix(virtualmodels): keep disabled legacy failover rules as disabled virtual models

SantiagoDePolonia:
The legacy `failover_rules` migration skipped rows with `enabled=0` but still deleted them and dropped the table, so a rule an operator had switched off in the dashboard lost its fallback list permanently on upgrade. Found by Greptile on #799.

A disabled dashboard rule is now converted the same way a primary listed in `FAILOVER_DISABLED_MODELS` already is: as a disabled failover virtual model that keeps its targets and can be re-enabled from the dashboard.

User-visible impact: after upgrading, disabled legacy rules appear as disabled virtual models instead of vanishing.

## Summary by CodeRabbit

* **New Features**
  * Legacy failover mappings disabled in the dashboard are now migrated as disabled virtual models.
  * Fallback targets remain configured but inactive until the mapping is re-enabled.

* **Documentation**
  * Updated migration guidance to explain how disabled failover mappings are handled.

### DIFF
--- a/docs/features/failover.mdx
+++ b/docs/features/failover.mdx
@@ -109,9 +109,10 @@ converted automatically:
   the setting is kept. A primary that already has a redirect, a path-scoped
   setting, or is disabled is logged and skipped, and the store is kept until
   it is resolved — add its fallbacks as targets of that virtual model with
-  the failover strategy, then delete the row. A primary listed in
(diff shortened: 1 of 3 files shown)


SantiagoDePolonia:
Stale

PR #815, opened 2026-08-31
fix(modeldata): resolve metadata for models with a routing suffix

apovalixin:
## Problem

OpenRouter-style variant suffixes — `:floor`, `:free`, `:nitro` — select how a
request is routed or billed, not which model answers it. The model registry
carries an entry for the base ID only, so a model requested under a suffixed ID
is published with no metadata at all: no display name, no context window, no
capabilities.

Seen on a running instance, where two entries of the same catalog sit next to
each other:

| Model | Context window | Display name |
|-------|----------------|--------------|
| `openrouter/openai/gpt-5.6-luna-pro` | 1050000 | OpenAI: GPT-5.6 Luna Pro |
| `openrouter/openai/gpt-5.6-luna-pro:floor` | — | — |

An alias pointing at the `:floor` variant inherits the empty record, so clients
that pick a model by its advertised context window cannot see one.

## Change

`resolveEntries` gains a final step that retries the lookup against the base ID,
mirroring the existing release-date fallback right above it. It runs last, so an
explicit registry entry for a suffixed ID (`deepseek/deepseek-r1:free` and the
like) still wins. A colon that belongs to a path rather than to a variant —
`host:8080/model` — is left alone, as is a leading or trailing colon.

## Tests

- `TestResolve_RoutingSuffixFallback` — `:floor`, `:nitro` and `:free` variants
  resolve to the unsuffixed entry.
- `TestResolve_RoutingSuffixDoesNotShadowExplicitEntry` — an explicit entry for
  the suffixed ID keeps priority.
- `TestStripRoutingSuffix` — boundary cases for the separator.

`go test ./...` on this branch: 82 packages pass. Three fail (`config`,
`internal/admin/dashboard`, `internal/server`) — they fail identically on the
unmodified base commit in the same environment, so they are unrelated to this
change.

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Model metadata now resolves correctly for supported routing suffixes such as `:floor`, `:free`, and `:nitro`.
  * Routing suffix fallback works alongside release-date variants.
  * Explicit entries with suffixed model IDs continue to take precedence.
  * Unknown, malformed, and path-based suffix formats are handled safely without incorrectly borrowing metadata.

### DIFF
--- a/internal/modeldata/merger.go
+++ b/internal/modeldata/merger.go
@@ -51,6 +51,17 @@ func resolveEntries(list *ModelList, providerType string, modelID string) (*Mode
 	return resolveRoutingSuffixAlias(list, providerType, modelID)
 }
 
+// routingSuffixes are the known routing/pricing variants a model ID may carry
+// after a colon. The set is closed on purpose: an arbitrary ":tail" can be
+// provider syntax with a different meaning — Bedrock encodes a model version
+// as "...-v1:0" — and stripping it would attach a neighbouring entry's
+// metadata to a distinct model.
+var routingSuffixes = map[string]struct{}{
+	"floor": {},
+	"free":  {},
+	"nitro": {},
+}
+
 // resolveRoutingSuffixAlias retries the lookup without a trailing routing
 // suffix such as ":floor", ":free" or ":nitro". The suffix picks how a request
 // is routed or billed, not which model answers it, and the registry carries an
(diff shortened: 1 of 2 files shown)


SantiagoDePolonia:
@apovalixin Good targeted increment! Thank you!

Feel free to share any feedback anytime about GoModel.

PR #821, opened 2026-08-31
fix(routing): let the route selector decide session affinity

SantiagoDePolonia:
## The bug

A session-affine virtual model pinned its target on the session's first request and then **skipped the strategy entirely** for as long as the pin lived. Under `adaptive` that took the route selector out of the path: it was consulted once per session, never again.

It survived review because the pin's viability test is *candidate membership*, and a target serving `429`s or timing out keeps passing it — it is still configured, still catalog-supported, still has rate-limit capacity. Only the selector knows it has been backed off. `ext/route.go` even documented the behaviour as intended:

> Session affinity is enforced by core before the selector runs; the ID is provided for observability only.

The consequence: an agent session — one session id, thousands of requests — pins on request one and then rides a failing target for the pin's whole lifetime. Failover keeps answering the client with `200`s, so it surfaces only as latency and wasted upstream spend.

## The change

Core hands the pin to the selector as `ext.RouteRequest.SessionTarget` and records whatever it answers, instead of pre-empting it.

- A selector that **declines** keeps core's affinity exactly as before, so declining never costs a session its pin.
- Every other strategy (`round_robin`, `cost`, `failover`) is untouched.
- With no selector installed, `adaptive` behaves as it always did.
- `stickySessions.pin` skips the map-wide prune when overwriting an existing key: the adaptive path records a choice on every request of every pinned session, and an O(n) sweep under the shared lock at that rate is a contention point, not housekeeping.

## Measured

Against three mock providers, `429`ing the pinned target:

| | before | after |
|---|---|---|
| `Select()` calls per 10 requests | 0 | 10 |
| hits on the failing target, batch 1 | 40 | 4 |
| batches 2–6 | 40 each | 0 |

Invariants checked in the other direction, because this change could plausibly have broken affinity outright:

- healthy session, 30 requests → 1 target, 30/30
- 30 distinct sessions → still spread across targets
- pinned target slowed to 400 ms → stays, 15/15 (slow is not unhealthy)

## Tests

Nine new tests. The three that encode the bug were verified to **fail against the pre-fix behaviour** and pass after; the rest are invariants that hold both ways.

This gap is why the bug shipped: both layers' unit tests passed, and only their interaction was broken.

## Not fixed here

A target that degrades to high latency but keeps returning `200`s never enters cooldown, so a pinned session stays on it. Closing that means inventing a "slow enough to evict" threshold — a design decision rather than a bug fix, so it is deliberately left out.

## Downstream

GoModel Pro's selector needs this `ext` field; its build fails against `v0.1.84`. Pro PR is open as a draft, blocked on this merging and being released.

https://claude.ai/code/session_013SNcqXgyv3sPZEtxSSMCYk

## Summary by CodeRabbit

* **New Features**
  * Adaptive routing now reevaluates session-pinned targets on every request.
  * Sessions can move automatically from failing or backed-off targets to healthy ones, while remaining pinned to slow but available targets.
  * Added session-affinity metrics for tracking honored and released pins.
  * Adaptive routing also applies to requests without an existing session pin.

* **Documentation**
  * Added guidance on adaptive session keeping, health handling, failover behavior, and related metrics.

### DIFF
--- a/docs/features/session-keeping.mdx
+++ b/docs/features/session-keeping.mdx
@@ -71,6 +71,15 @@ saturated, the strategy picks again and the session re-pins — a session is
 never glued to a dead target, and the all-saturated honest-429 behavior is
 unchanged.
 
+Under the `adaptive` strategy the pin is handed to the route selector, which
+decides on every request whether the session keeps it. "Unavailable" then
+also covers a target the selector has backed off — one serving 429s or
+failing repeatedly — which stays in the target list and so looks fine to the
+checks above. The session moves to a healthy target and re-pins there. A
+target that is merely *slow* still keeps its session: moving it would throw
+away the provider prompt cache the pin exists to protect. See
+[Intelligent Routing](/pro/intelligent-routing).
+
 Affinity is on by default and can be disabled per redirect:
 
 ```yaml
--- a/docs/features/session-keeping.mdx
+++ b/docs/features/session-keeping.mdx
@@ -72,7 +72,10 @@ never glued to a dead target, and the all-saturated honest-429 behavior is
 unchanged.
 
 Under the `adaptive` strategy the pin is handed to the route selector, which
-decides on every request whether the session keeps it. "Unavailable" then
+decides on every request whether the session keeps it — whenever more than
+one target is viable. (With a single viable target there is nothing to
+choose, so the selector is skipped and the redirect behaves like a plain
+alias.) "Unavailable" then
 also covers a target the selector has backed off — one serving 429s or
 failing repeatedly — which stays in the target list and so looks fine to the
 checks above. The session moves to a healthy target and re-pins there. A
--- a/docs/pro/intelligent-routing.mdx
+++ b/docs/pro/intelligent-routing.mdx
@@ -171,7 +171,9 @@ The objective decides how price enters the weight:
 
 With [Session keeping](/features/session-keeping) on (the default), the
 gateway hands the selector the target already serving each session and lets
-it decide, request by request, whether the session keeps it. The pin holds
+it decide, request by request, whether the session keeps it — whenever more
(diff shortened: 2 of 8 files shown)


SantiagoDePolonia:
Both concurrency findings were real and reproduced with tests that fail against `40cd79f6`. Fixed in `40211a22`.

**Concurrent session split (P1 / CodeRabbit + Greptile).** `sticky.pin` is replaced by `repin(source, session, observed, chosen)`, a compare-and-set: it commits against the pin the selector was actually shown, and when another request pinned something else meanwhile, that pin wins and is returned so both requests leave together. This restores for the adaptive path the atomic recheck `sticky.resolve` already gave the other strategies. Covered by `TestSticky_AdaptiveConcurrentFirstRequestsAgree` (no prior pin) and `TestSticky_AdaptiveConcurrentRepinsAgree` (deliberate replacement of an existing pin) — both use a barrier selector that holds every concurrent call until they have all observed the same pin, then hands each a different valid answer.

**Rotation slot consumed on the fallback path (P2 / Greptile).** Selector consultation moves out of `pick` into `selectorChoice`, so the affinity path can tell a decline from an answer and returns a viable pin directly instead of building a round-robin fallback it is about to discard. `TestSticky_AdaptiveDeclineDoesNotAdvanceRoundRobin` asserts the counter is unchanged across all three unusable-selector modes — declines, answers outside the pool, and panics. It failed `1 -> 5` before.

**Docs (CodeRabbit).** Correct — `adaptiveTarget` is bypassed at a single viable target. Both passages now qualify "every request" with the multiple-viable-target condition.

Re-verified end to end after the restructure: hits on a 429ing pinned target still go 40/batch → 4 once → 0, sustained over 60 requests; healthy sessions still stick 30/30; a target merely slowed to 400 ms still keeps its session 15/15. Full package green under `-race -count=3`.

PR #822, opened 2026-08-31
feat(openai): detect speech-to-text-only providers without /v1/models

### LINKED ISSUE #819: Issue: STT provider is not being recognised
STT in openai is unfortunately not clearly defined. There is no requirement from a provider to present any model, it may present `/health` and `/v1/audio/transcriptions` endpoints only.

IMHO GoModel should actively scan a provider for `/v1/audio/transcriptions` and then assume that it's an STT provider.

SantiagoDePolonia:
Fixes #819.

OpenAI-compatible STT servers (whisper.cpp server, faster-whisper wrappers) may expose only `/v1/audio/transcriptions` — no `GET /v1/models`. Today such a provider fails model discovery and never becomes routable.

For providers of `type: openai`, when `GET /models` reports the route missing (404/405/501) or returns an empty list, GoModel now probes `POST /audio/transcriptions` with an empty body. If the upstream rejects it with a request-level error (400/413/415/422), the provider advertises a synthesized `whisper-1` model with the `audio_transcription` mode, so OpenAI SDK clients (which default to `whisper-1`) route to it with zero configuration.

Behavior notes:

- Detection is skipped when the provider declares `models:` — with the default configured-models mode (`fallback`) a successfully synthesized inventory would otherwise silently override the operator's list.
- Auth/availability failures on `/models` are never treated as "endpoint missing", and a 401/403 on the probe is not accepted as proof the endpoint exists (middleware that authenticates before route matching emits those for absent routes too), so bad credentials or outages surface as fetch errors instead of a phantom model.
- Other wrappers of the shared OpenAI-compatible transport (Azure, OpenRouter, chat-centric providers) leave the flag off; nothing changes for them or for api.openai.com, where `/models` always answers.

Docs: `config.example.yaml` and the Audio API page document the detection and the `models:` override for servers that require their own model name.

## Summary by CodeRabbit

- **New Features**
  - Added automatic support for OpenAI-compatible speech-to-text servers that do not provide a model list.
  - These servers now advertise the `whisper-1` transcription model when detected.
  - Custom model configurations take precedence over automatic detection.
- **Documentation**
  - Added configuration examples and guidance for speech-to-text-only servers.
- **Bug Fixes**
  - Improved model discovery when compatible servers omit or return an empty model catalog.

### DIFF
--- a/config/config.example.yaml
+++ b/config/config.example.yaml
@@ -546,6 +546,15 @@ providers:
   #   base_url: "https://api.example.com/v1"
   #   api_key: "..."
 
+  # Speech-to-text-only servers that expose /v1/audio/transcriptions without
+  # GET /v1/models (whisper.cpp server, faster-whisper wrappers) are detected
+  # automatically and advertise a "whisper-1" transcription model. Declare
+  # `models:` instead when your server requires its own model name.
+  # local-whisper:
+  #   type: openai
+  #   base_url: "http://localhost:8000/v1"
+  #   api_key: "unused"  # any non-empty value if the server ignores auth
+
   # Example: LM Studio. Despite being "Ollama-like", LM Studio speaks the
   # OpenAI-compatible API (/v1/chat/completions, /v1/embeddings) and has NO
   # native Ollama API. Configure it as "openai" (or "vllm"), NOT "ollama" —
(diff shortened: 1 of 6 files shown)


SantiagoDePolonia:
Not needed - the issue https://github.com/ENTERPILOT/GoModel/issues/819 was closed by a different approach.

PR #868, opened 2026-09-02
feat(auth): scope admin API and lifecycle objects to the key's user path

SantiagoDePolonia:
## Summary

A managed API key's bound `user_path` is now its **access scope**. A key bound to `/team/alpha` can only administer and address `/team/alpha` and its descendants. The master key, keys without a `user_path`, and keys bound to `/` stay global. The `X-GoModel-User-Path` header never widens a scope.

This fixes two related problems:

- A managed key with `dashboard_access=true` was a full admin regardless of its `user_path`.
- Responses, conversations, batches, and files persisted `user_path` but were looked up by ID only, so any credential that knew an ID could read, mutate, or delete another tenant's object.

## User-visible impact

- **Admin API and dashboard.** New `GET /admin/access` reports the caller's scope. Usage, audit (including stats), API keys, users, budgets, and rate limits are filtered to the scope; an omitted `user_path` filter means the scope root. Gateway-wide endpoints (providers, credentials, runtime settings, tagging, virtual models, workflows, guardrails, MCP servers, pricing overrides, cache overview, live logs, throughput, pricing recalculation, reset-all) answer `403 admin_scope_denied` for scoped admins. Caller-named paths outside the scope answer `403 user_path_out_of_scope`. Objects addressed by ID outside the scope answer `404`, never `403`.
- **`/v1` lifecycle objects.** Get, list, update, cancel, and delete on responses, conversations, batches, and files check the stored `user_path` against the scope. Tracked objects outside the scope return `404` without falling through to the provider. Untracked IDs keep the provider fallback (documented).
- **Dashboard.** Scoped admins see a "Scoped to /team/alpha" indicator, global-only pages and actions are hidden, and editors default their user path to the scope root.
- **Behavior change.** Existing dashboard keys that carry a `user_path` lose gateway-wide admin rights. Rows with an empty `user_path` are visible to global credentials only.

## Docs

`docs/features/user-path.mdx` (Access scope, Object ownership), `docs/advanced/admin-endpoints.mdx` (Scoped admin access, `GET /admin/access`), regenerated `docs/openapi.json`.

## Verification

`make lint` clean, `go test ./...` green, dashboard `npm run check` and `npm test` green.

https://claude.ai/code/session_01V7kPQrb4xiNfTW25B5QmbK

## Summary by CodeRabbit

* **New Features**
  * Added scoped administrative access based on user paths.
  * Added `GET /admin/access` to report the current administrative scope.
  * Added user-path filtering for audit statistics, administrative listings, files, and batches.
  * Updated the dashboard to display scope information and hide unavailable pages.
  * Added persistent user-path ownership and access controls for responses, conversations, batches, and files.

* **Bug Fixes**
  * Out-of-scope objects now return not-found responses.
  * Gateway-wide administrative actions are blocked for scoped credentials.
  * User-path headers no longer expand a credential’s assigned scope.
  * New keys, users, budgets, and rate limits default to the permitted path.


SantiagoDePolonia:
Addressed the review findings in 6516acf:

- **Scoped batch listing (Greptile P1, CodeRabbit):** batches now carry a `user_path` column (SQL: added via `AddColumns` and backfilled from the JSON payload on start; Mongo: field plus backfill). Scoped `List` filters in the store, so paging is exact, `has_more` is honest, and a foreign cursor is reported as missing. The capped scan and its nil-row cursor read are gone.
- **Scoped file listing (CodeRabbit):** `GET /v1/files` for a scoped credential now lists the caller's tracked files from the gateway's ownership records (`filestore.Store.List` with subtree, provider, and purpose filters), so no provider paging cap can hide owned files. `GetMany` was removed.
- **Unmapped file IDs (CodeRabbit IDOR):** scoped callers get 404 for any file the gateway does not track, with no provider enumeration. The same rule now applies to untracked response IDs. Global credentials keep the provider fallback; documented in `docs/features/user-path.mdx`.
- **`ErrorTypePermission` default status** mapped to 403.
- **OpenAPI:** scoped admin endpoints document the 403 response; spec regenerated.
- **Tests:** the auth scope test now seeds an ambient scope before the bearer replaces it.
- **Dashboard:** access scope reloads after a stale response, budget and rate-limit editors force `user_path` when the credential turns out to be scoped and reject subjects outside the scope before submitting, gateway-wide settings panels wait for the scope to load, and the Polish message uses the established term.

Not changed: `inherited_from` on `/admin/users` still lists ancestor policy paths above the scope root. Those paths are prefixes of the caller's own path, so they reveal nothing new, and the `restricted` flag is only explainable with them.

SantiagoDePolonia:
Follow-ups in 504244c:

- **CodeQL (Mongo query from user input):** the subtree filters in both Mongo stores now use byte-range bounds (`user_path >= path+"/"` and `< path+"0"`, the same trick the audit reader uses) instead of a regex built from the scope path.
- **Backfill on start-up (CodeRabbit):** the batch `user_path` backfill runs in chunks of 500 rows, pauses with a warning on any error instead of failing store construction, and is resumable because unmigrated rows keep a NULL column (SQL) or lack the field (Mongo). Mongo runs it under its own five-minute deadline rather than the index deadline. Batches are low-volume per gateway, so the synchronous pass stays short; until it completes, scoped listings simply do not show the unmigrated rows.
- **Pending scope in editors:** budget and rate-limit submits now await `access.ensureLoaded()` before validating against the scope, so a key change mid-edit cannot slip a gateway-wide rule through.
- **Polish message:** refers to `{root}` explicitly.

SantiagoDePolonia:
One more correction in the latest commit, found while checking compatibility with GoModel Pro SSO:

An SSO session always carries a personal user path (`/users/<email>` by default) and grants dashboard access from admin-group membership. Deriving the access scope from that path would have confined every SSO admin to their own identity path and broken the Pro dashboard. Extension identities now stay global unless the extension sets the new `ext.Authentication.AccessScope` field; their `UserPath` remains an attribution identity only. Managed API keys are unchanged: their bound `user_path` is their scope. Docs updated accordingly.

SantiagoDePolonia:
Reverted the previous commit's `ext.Authentication.AccessScope` field. It was a second policy axis, and defaulting extension identities to global only preserved a behavior that was not right.

One rule now holds for every credential: the user path is the node you act as, and you can see that node and everything below it. An SSO session bound to `/users/alice` administers that node only. An extension hands out gateway admins by binding their sessions to `/` and team admins by binding them to the team root, the same way you would create an API key at that path. Docs updated.

GoModel Pro needs a matching change before it bumps to a core release with this PR: bind sessions of `admin_groups` members to `/` (or a configured admin root) instead of `/users/<email>`.

PR #877, opened 2026-09-03
feat(config): add GOMODEL_OFFLINE switch and local file model catalog source

SantiagoDePolonia:
## Summary

Two things an air-gapped operator needs from core, plus a doc correction.

**`GOMODEL_OFFLINE=true`** (`offline: true` in `config.yaml`) is one switch that disables every outbound call the gateway makes on its own: the update check and the remote model catalog download. It is applied after every other config source, so neither `config.yaml` nor `GOMODEL_VERSION_CHECK_ENABLED=true` can re-enable a call underneath it. Calls to configured providers and operator-declared endpoints (OTLP, MCP upstreams, vector stores) are untouched. Startup logs the mode. Default: `false`, nothing changes for existing deployments.

**`MODEL_LIST_URL` accepts a local file.** A bare path (`/etc/gomodel/models.json`) or `file://` URL is read on startup and every cache refresh, and re-parsed only when its content changes (validator is a SHA-256 of the file, reusing the existing ETag path so `NotModified` short-circuits exactly like a 304). A local file involves no network request, so it stays active under `GOMODEL_OFFLINE=true`. This lets air-gapped sites keep pricing and budgets without running an HTTP mirror.

**Doc fix.** `docs/guides/production.mdx` claimed "no update check", but the version check ships enabled by default. The air-gap section now lists both outbound calls with their defaults, the single switch, and the three ways to keep pricing offline.

## Docs

- `docs/guides/production.mdx`: rewritten "Air-gapped and offline deployments" section, updated checklist.
- `docs/advanced/model-metadata.mdx`: file source under "Offline behavior".
- `docs/advanced/version-awareness.mdx`: pointer to the offline switch.
- `docs/advanced/configuration.mdx`: `MODEL_LIST_URL` and `GOMODEL_OFFLINE` rows.
- `.env.template`, `config/config.example.yaml`.

## Testing

- `config`: offline drops remote and mirror URLs, keeps file and bare-path sources, wins over an explicit `version_check.enabled: true`; default stays online; `IsLocalModelListSource` table.
- `internal/modeldata`: local file read for path and `file://`, NotModified on unchanged content, re-read on change, missing file and invalid JSON errors, `localPath` table.
- Full `make test-race` and `make lint` pass via pre-commit.

## Summary by CodeRabbit

* **New Features**
  * Added offline mode to disable gateway-initiated update checks and remote model catalog downloads.
  * Added support for loading model catalogs from local files, including `file://` paths, with change detection during refresh.
  * Preserved configured providers and local catalog access while offline mode is enabled.
  * Added startup messaging that reports offline mode and catalog status.

* **Documentation**
  * Documented offline configuration, local catalogs, per-call controls, and air-gapped deployment options.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -21,6 +21,13 @@
 # values stay fatal in either mode.
 # CONFIG_STRICT=true
 
+# Air-gapped installs: one switch that disables every outbound call the gateway
+# makes on its own -- the update check (GOMODEL_VERSION_CHECK_*) and the remote
+# model catalog download (MODEL_LIST_URL over HTTP). Configured providers and
+# endpoints you declare yourself (OTLP, MCP upstreams, vector stores) are not
+# affected, and a catalog given as a local file path keeps working.
+# GOMODEL_OFFLINE=false
+
 # Tagging based on headers: label every request from the listed headers (numbered
 # from 1). Labels are recorded in usage tracking and audit logs. A header value can
 # carry several labels split by the delimiter (default: ","). The optional prefix is
@@ -219,8 +226,9 @@
 
 # External model metadata registry (provides pricing, capabilities, context window, etc.)
 # Default: ENTERPILOT/ai-model-list on GitHub. Point this at an internal mirror,
-# or set MODEL_LIST_URL=off to disable the download entirely for fully
-# air-gapped installs -- models then keep provider-reported, configured,
+# at a local file (/etc/gomodel/models.json or file:///etc/gomodel/models.json,
+# re-read on each cache refresh), or set MODEL_LIST_URL=off to disable the
+# catalog entirely -- models then keep provider-reported, configured,
 # heuristic-derived, and any previously cached catalog metadata, so declare
 # pricing in config.yaml if you need cost tracking and budgets. (Setting it
 # to an empty string does NOT disable the fetch: empty env values are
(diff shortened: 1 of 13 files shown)


SantiagoDePolonia:
Addressed the bounded-read finding (Greptile, CodeRabbit): `readLocal` now rejects an oversized catalog from `Stat` before allocating, and reads through `io.LimitReader(maxBodySize+1)` as a backstop if the file grows between stat and read. New `TestFetchIfChanged_LocalFileOversizedIsRejectedBeforeAllocation` uses an 80 MiB sparse file and asserts total allocation stays under the 10 MiB limit.

SantiagoDePolonia:
@coderabbitai review

SantiagoDePolonia:
Addressed the blocking-path finding (CodeRabbit): `readLocal` now stats the path first and rejects anything that is not a regular file before `os.Open`, so a FIFO or device at `MODEL_LIST_URL` fails fast instead of hanging startup. Covered by `TestFetchIfChanged_LocalFIFODoesNotBlock` and a directory case in the errors test.

SantiagoDePolonia:
Addressed both follow-ups (CodeRabbit):
- `readLocal` now opens with `O_NONBLOCK` and validates the descriptor it holds via `f.Stat()`, so a path swapped for a FIFO between check and open cannot block. Verified with `GOOS=windows go vet` that `syscall.O_NONBLOCK` resolves there too.
- The FIFO regression test moved to `fetcher_unix_test.go` behind `//go:build unix`; `syscall.Mkfifo` no longer appears in a cross-platform file.

PR #882, opened 2026-09-04
fix(server): log unhandled errors and panics and answer in the gateway envelope

### LINKED ISSUE #881: Admin API /admin/virtual-models and /admin/provider-credentials return HTTP 500 (empty body) since 0.1.84
### Summary
On latest release **v0.1.86**, GET `/admin/virtual-models` and GET `/admin/provider-credentials` both return **HTTP 500** with an **empty body**. The dashboard's "Virtual Models" and "Providers" panels therefore fail to load. The API key auth works fine (401 -> 500 boundary is clean); the server crashes during response serialization.

### Environment
- GoModel version: **v0.1.86** (commit 7c934f49, built 2026-09-02T19:13:49Z)
- OS: Windows (local single-node)
- Storage: SQLite
- Auth: proper `Authorization: Bearer <master-key>`

### Reproduction
1. Start gomodel v0.1.86.
2. `curl -H "Authorization: Bearer <master-key>" http://127.0.0.1:8080/admin/virtual-models`
3. Receive `HTTP 500` with **empty response body**.
4. Same for `GET /admin/provider-credentials`.

**Without auth**, the same endpoint returns `401 {"error":{"message":"missing credentials: ..."}}` — so authentication is correctly enforced and the crash happens after auth, during handler/serialization.

### Expected
HTTP 200 with the list of virtual models / provider credentials.

### Impact
- Dashboard web UI: "Virtual Models" and "Providers" panels fail to load (blank/error).
- **Actual model routing is NOT affected**: `GET /v1/models` returns 200 with all models; virtual models (e.g. `mycoding`, `myvision`) still route requests successfully.
- This is purely an admin/management-plane issue, not a data-plane issue.

### Notes / related
- The 0.1.84 changelog contains a **breaking change**: *"Admin API: virtual-model responses now use a targets[] array instead of separate provider/model fields (#788)"*. Our DB data is already the new `targets[]` format, yet GET /admin/virtual-models still 500s — suggesting a handler regression in the 0.1.84+ admin API.
- The server is launched with stderr discarded, so no panic stack is visible; the 500 has no JSON error body.

### Suggested
Please check the admin handlers for `/admin/virtual-models` and `/admin/provider-credentials` for a regression introduced in 0.1.84 (serializing `targets[]` or provider credential rows). Happy to provide more detail / debug logs if useful.

---
*Environment details provided by s-edi (Windows) on 2026-09-04.*

SantiagoDePolonia:
Fixes #881 (diagnosability). Any error that escaped a handler — a recovered panic or a response serialization failure — was answered with echo's bare `{"message":"Internal Server Error"}` and logged nowhere: `middleware.Recover` in echo v5 only converts the panic into an error, and the default `HTTPErrorHandler` writes the body without logging. That is what the reporter hit on `GET /admin/virtual-models` and `GET /admin/provider-credentials`: a 500 with no gateway error body and no trace of the cause, even with logs captured.

This installs a centralized `HTTPErrorHandler` that:

- logs `unhandled request error` at ERROR level with method, path, request id, and, for panics, the panic value and stack (`stack` attribute);
- renders the response in the gateway error envelope (`{"error":{"type":"internal_error","message":"an unexpected error occurred",...}}`), without leaking the internal error text;
- leaves echo's own status errors (body limit 413, method not allowed 405, ...) to the default handler so their status and body are unchanged.

I could not reproduce the underlying failure of the two admin endpoints on macOS against a real SQLite database, and neither handler changed between 0.1.83 and 0.1.86. With this change the reporter's log will show the actual panic or serialization error (for example a stored timestamp outside the JSON-encodable year range), so the root cause can be fixed in a follow-up.

https://claude.ai/code/session_01NEgRAh914CS85pUNh92RBh

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected server errors, including panics and response failures.
  - Errors now return a consistent gateway error response instead of a generic error message.
  - Added safeguards to avoid overwriting responses that have already been sent.
  - Preserved existing HTTP error statuses and response bodies for recognized client errors.

- **Monitoring**
  - Unexpected errors are now logged with request details and diagnostic information, including stack traces for recovered panics.

### DIFF
--- a/internal/server/http.go
+++ b/internal/server/http.go
@@ -145,6 +145,10 @@ func New(provider core.RoutableProvider, cfg *Config) *Server {
 		JSONSerializer: goJSONSerializer{},
 	})
 	e.Logger = slog.Default()
+	// Errors that escape the handler chain (recovered panics, response
+	// serialization failures) are logged and rendered in the gateway envelope
+	// instead of echo's silent {"message": "Internal Server Error"}.
+	e.HTTPErrorHandler = unhandledErrorHandler(e.HTTPErrorHandler)
 	basePath := configuredBasePath(cfg)
 	if basePath != "/" {
 		e.Pre(stripBasePathMiddleware(basePath))
--- a/internal/server/unhandled_error.go
+++ b/internal/server/unhandled_error.go
@@ -1,0 +1,90 @@
+package server
+
+import (
+	"errors"
+	"log/slog"
+	"net/http"
+
+	"github.com/labstack/echo/v5"
+	"github.com/labstack/echo/v5/middleware"
+
+	"github.com/enterpilot/gomodel/internal/core"
+)
+
+// unhandledErrorHandler is the echo HTTPErrorHandler: it receives every error
+// that escapes the handler chain without being rendered — a panic recovered by
+// middleware.Recover, a response serialization failure, or an echo status
+// error raised by a built-in middleware (body limit, method not allowed).
+//
+// echo's default handler answers with a bare {"message": "Internal Server
+// Error"} body and logs nothing, so a panicking admin handler left operators
+// with a 500 and no trace of the cause. Unexpected errors are logged with the
+// request path and, for panics, the stack, and rendered in the gateway error
+// envelope. echo's own HTTPErrors keep their status and body through fallback.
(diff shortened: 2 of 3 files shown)


SantiagoDePolonia:
Redundant - #882 is a better fix

PR #886, opened 2026-09-04
fix(dashboard): stop audit session list and interactions drawer from jumping during live streams

SantiagoDePolonia:
## What & why

Two audit-log dashboard UX issues where the view jumps while records stream in real time.

### 1. Session list jumps when live records arrive (grouped mode)

The gateway emits `audit.started` before session detection runs, so a new live request arrives with no `session_id`. The old code inserted it as its own singleton thread at the top of the list, then re-folded it into its real thread milliseconds later when the session update landed — two list mutations per request, shifting the whole list twice.

Now a sessionless live row is held off-screen briefly and placed exactly once, when its outcome is known: the session update arrives (folds straight into its thread), the request settles or a model/workflow is stamped with still no session (placed as its own thread), or a short grace period elapses. Thread folding is also timestamp-aware, so a late-arriving request slips in as a child rather than displacing a newer head. The normalized record cache still receives every event, so an open interactions drawer keeps following.

### 2. Interactions drawer jumps while following a conversation

Two causes: the follow-scroll used `scrollIntoView`, which scrolls every scrollable ancestor (dragging the audit list behind the drawer), and it re-scrolled to the bottom on every record change (yanking a reader who had scrolled up).

The scroll helpers now move the drawer's own container directly, and the follow pin is tracked from the container's scroll events — scrolling up unpins, scrolling back down re-pins. A live flush re-hydrates the transcript without forcing the view down. Separately, the highlighted open-interactions row now uses a box-shadow ring at constant border width instead of a thicker border, so rows below it no longer shift a pixel.

## User-visible impact

- Unfolded session threads stay put as live requests stream in.
- The interactions drawer follows new content only when pinned to the bottom; a reader scrolled up stays where they are.

## Testing

- `npm run check` — 0 errors
- `npm test` — 588 pass / 0 fail (new tests for the hold lifecycle and the drawer pin/anchor helpers)
- Verified live in the browser against a streaming gateway: head row held stable across 153 samples during a live arrival; drawer held position through a full stream + settle when scrolled up, and kept following when pinned.

Frontend only; no Go changes. The embedded dashboard bundle is rebuilt by CI.

https://claude.ai/code/session_01L8E6hziQAZjaGTch7E9tDa

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation drawer scrolling so viewing or streaming messages no longer unexpectedly moves the surrounding audit log.
  * Live conversation updates now continue following the latest content only when the reader is already near the bottom.
  * Selecting a message centers it within the conversation drawer.
  * Reduced visual row shifting when switching between interaction drawers.
  * Improved grouping of live audit entries by briefly waiting for session information before placing entries in the list.

### DIFF
--- a/web/dashboard/src/pages/audit-logs/AuditEntryRow.svelte
+++ b/web/dashboard/src/pages/audit-logs/AuditEntryRow.svelte
@@ -98,13 +98,17 @@
       border-color 0.12s ease-out;
   }
 
+  /* A ring rather than a thicker border: a border-width change shifts every
+     row below by a pixel each time the Interactions drawer moves to another
+     entry. */
   .audit-entry-interactions-open {
-    border-width: 2px;
     border-color: color-mix(
       in srgb,
       var(--prompt-cache-color) 38%,
       var(--border)
     );
+    box-shadow: 0 0 0 1px
+      color-mix(in srgb, var(--prompt-cache-color) 38%, var(--border));
   }
 
   .audit-entry-details {
--- a/web/dashboard/src/pages/audit-logs/ConversationDrawer.svelte
+++ b/web/dashboard/src/pages/audit-logs/ConversationDrawer.svelte
@@ -131,7 +131,7 @@
   }
 
   function scrollToConversationMessage(direction) {
-    const content = document.getElementById("interactions-drawer-content");
+    const content = drawer.conversationContentEl;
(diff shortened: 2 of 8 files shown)


SantiagoDePolonia:
It's breaking the current behavior...

PR #893, opened 2026-09-05
fix(gemini): preserve function-call thought signatures across the round trip

SantiagoDePolonia:
Gemini returns a thoughtSignature on the part carrying a functionCall and
Gemini 3 rejects the next turn with HTTP 400 when a replayed functionCall part
has lost it. The signature was dropped converting the native response to
OpenAI tool calls and never restored converting history back, so multi-step
agent runs failed once a tool call reached the history.

It now travels on the tool call as extra_content.google.thought_signature, the
member Google's own OpenAI-compatible endpoint uses, through the buffered and
streamed chat surfaces and the Responses API in both directions. A flat
thought_signature is accepted on input too.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_011CneM5UC4k3TikAMgEvm8o

## Summary by CodeRabbit

- **New Features**
  - Gemini tool calls now preserve thought signatures across requests, responses, streaming, and multi-step tool interactions.
  - OpenAI-compatible clients can provide and receive signatures through `extra_content.google.thought_signature` and supported flat formats.
  - Gemini 3 tool-call signatures are retained when replaying assistant messages.

- **Documentation**
  - Added guidance on Gemini tool calling and thought signatures, including a link to Google’s reference documentation.

- **Tests**
  - Added coverage for native, streaming, Responses API, and end-to-end Gemini 3 signature preservation.

### DIFF
--- a/docs/providers/gemini.mdx
+++ b/docs/providers/gemini.mdx
@@ -89,6 +89,38 @@ Flash reject `minimal`, so GoModel sends the lowest level those models accept
 instead of surfacing Google's 400. An explicit
 `extra_body.google.thinking_config` is always forwarded unchanged.
 
+## Tool calling and thought signatures
+
+Gemini returns a `thoughtSignature` alongside the function call it emits, and
+Gemini 3 requires that signature back, unchanged, on the same function call
+when the conversation history is sent again. A missing one fails the whole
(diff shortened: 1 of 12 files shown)


SantiagoDePolonia:
#894 has a better fix

PR #928, opened 2026-09-08
fix(anthropic): carry thinking signatures back to the client

SantiagoDePolonia:
Anthropic signs every thinking block it returns and rejects a later turn that replays one without its signature. GoModel dropped the signature on the way out, so the standard agent loop — echo the assistant turn back — always failed on the second request:

```
POST /v1/messages  thinking.enabled  ->  {"type":"thinking","thinking":"..."}   (no signature)
replay that content                  ->  400 messages.1.content.0.thinking.signature: Field required
```

## User-visible impact

Multi-turn extended thinking now works on every dialect routed to Anthropic, tool-use loops included.

The blocks travel out the same way they already travel in, on `extra_content.anthropic.thinking_blocks` (the mechanism added in #895):

- **`/v1/messages`** renders the native `signature` and `data` fields on the content blocks, because that is what an Anthropic client echoes back. Streaming emits a `signature_delta` inside the thinking block before it closes, as Anthropic does. `extra_content` is never leaked into this dialect.
- **Chat Completions** carries the vendor object on the assistant message, next to the unchanged `reasoning_content`. Streaming emits it once the thinking block completes, before any text, cumulatively.
- **Responses** carries it on the `reasoning` output item and restores it from an echoed one.
- Synthesized streams (`SynthesizeChatStream`) carry the message's replay state too, so an un-streamed response is still replayable.

`redacted_thinking` has no readable text, so `extra_content` is the only carrier for it outside the Messages dialect.

## Provider-specific notes

- The Anthropic **Responses** path surfaced no reasoning at all — no `reasoning_content`, no reasoning item. It does now. Streaming `/v1/responses` still emits none (the converter never wired up the shared reasoning state); documented as a limitation rather than fixed here, to keep this change focused.
- Routing a history to another provider drops `extra_content.anthropic` as before; verified live that an Anthropic thinking turn is accepted by OpenAI.
- `SynthesizeChatStream` still drops **tool call** `extra_content` (a separate, pre-existing Gemini gap).

## Tests

Written before the fix, all failing at the time:

- `tests/contract/extra_content_roundtrip_test.go` — response-side mirror of `TestAnthropicThinkingBlocksReplay`: a signed thinking block plus a redacted one must reach the client and come back verbatim on the upstream request through all three ingress dialects.
- Streaming coverage on both sides: `signature_delta` -> replay state (provider) and replay state -> `signature_delta` (egress), including cumulative values not re-emitting, and a guard that a stream with no thinking gains no `extra_content`.
- `FromChatResponse` table: signed blocks verbatim, fallback to `reasoning_content` for providers with no thinking protocol, malformed member falls back, another vendor's state is not thinking.

Verified live against the Anthropic API: two-turn thinking, a streamed turn reassembled and replayed, a tool-use agent loop with parallel calls, and all three dialects. The same script against the pre-fix build fails turn 2 with the 400 above.

`cmd/gomodel/docs/docs.go` also picks up the `/v1/auth/verify` route from #924, which was left stale there; that is regeneration, not a change in this PR.

## Summary by CodeRabbit

- **New Features**
  - Added an authentication verification endpoint that reports credential validity and authentication status.
  - Anthropic thinking and redacted-thinking blocks now preserve signatures and opaque data across requests, responses, streaming, and tool-use workflows.
  - Signed thinking blocks replay correctly across chat, Responses, and Anthropic Messages API formats.

- **Documentation**
  - Expanded API documentation for thinking-block fields, replay behavior, streamed deltas, and `extra_content`.
  - Clarified audit-log revision ordering and prompt-guardrail recording behavior.

### DIFF
--- a/cmd/gomodel/docs/docs.go
+++ b/cmd/gomodel/docs/docs.go
@@ -4335,6 +4335,37 @@ const docTemplate = `{
                 ]
             }
         },
+        "/v1/auth/verify": {
+            "get": {
+                "description": "Reports whether the presented credential authenticates against this gateway. Returns 401 when it does not. A gateway with no authentication configured has no credential to confirm and answers 200 with valid=false and method=none. Disabled unless AUTH_VERIFY_ENABLED is set.",
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "auth"
+                ],
+                "summary": "Verify an API key",
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/server.authVerifyResponse"
+                        }
+                    },
+                    "401": {
+                        "description": "Unauthorized",
+                        "schema": {
+                            "$ref": "#/definitions/core.OpenAIErrorEnvelope"
+                        }
+                    }
+                },
+                "security": [
+                    {
+                        "BearerAuth": []
+                    }
+                ]
+            }
+        },
         "/v1/batches": {
             "get": {
                 "produces": [
@@ -8706,6 +8737,10 @@ const docTemplate = `{
         "anthropicapi.ResponseContentBlock": {
             "type": "object",
             "properties": {
+                "data": {
+                    "description": "Data is the opaque payload of a redacted_thinking block.",
+                    "type": "string"
+                },
                 "extra_content": {
                     "description": "ExtraContent is provider replay state on a tool_use block; clients echo\nit back on the next turn (see core.ExtraContentField).",
                     "type": "object"
@@ -8718,6 +8753,10 @@ const docTemplate = `{
                     "additionalProperties": true
                 },
                 "name": {
+                    "type": "string"
+                },
+                "signature": {
+                    "description": "Signature authenticates a thinking block. Anthropic requires it back\nverbatim when the conversation continues, so clients must echo it.",
                     "type": "string"
                 },
                 "text": {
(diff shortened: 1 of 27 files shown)


SantiagoDePolonia:
All five review findings verified and fixed in 3e70380. Two were real bugs, not just test gaps.

**Stale reasoning replay state (`responses_input.go`) — real bug.** `flushPendingAssistant` returned early when `pendingAssistant` was nil, so the pending state survived. With `[reasoning, user, assistant]` the reasoning item's blocks were pinned to the later, unrelated assistant turn — exactly what Anthropic rejects. Both `pendingReasoning` and `pendingReasoningExtra` are now cleared on every flush: a reasoning item describes the turn that follows it, and once the history moves past that turn the state is spent. Test covers both the leak and the ordinary attach.

**`AssembleChatResponse` dropped `extra_content` — real bug.** This is the buffered plugin path (`plugin_stream.go` assembles, runs the plugin, re-synthesizes). My `SynthesizeChatStream` change made the synthesizer emit replay state, but the assembler never put it on the response, so a response-phase plugin stripped the signature from a turn it did not otherwise change. Now carried through, cumulative (last chunk wins), with a test.

**OpenAPI schema.** Confirmed, but `make swagger` does not fix it: the request `ContentBlock` schema is hand-written — and duplicated — in `tools/swagger-postprocess.mjs` and `tools/openapi-postprocess.mjs`, because of the recursive `content` union. Added `signature` and `data` to both and regenerated; `docs.go` and `openapi.json` now carry them. (`title` is also missing from that hand-written schema, unrelated to this PR.)

**Test strengthening (2 and 4).** `TestConvertAnthropicResponseToResponses_WithThinkingBlocks` now populates `Thinking`/`Signature` instead of `Text`, compares the full `thinking_blocks` payload, and adds a `redacted_thinking` case. The synthesized-stream test asserts `extra_content` precedes the first `content` delta.

Re-verified live against Anthropic after the changes: two-turn thinking, a tool-use agent loop with parallel calls, the Responses path, and cross-provider stripping to OpenAI.

SantiagoDePolonia:
Greptile's second finding (streamed replay state lost) is now fixed in 1a064a5 rather than documented.

The Anthropic Responses stream converter skipped `thinking_delta` and `signature_delta` outright, so a streamed `/v1/responses` turn carried no reasoning at all — not just no signature. It now mirrors the OpenAI converter: reserves the first output slot on the first thinking block, emits `response.reasoning_text.delta`, and closes the item when text or a tool call follows (`ResponsesOutputEventState` already had the whole reasoning lifecycle; it was simply never wired up here). The signature rides on the reasoning item as `extra_content`, so a streamed turn replays like every other surface. `SetReasoningExtraContent` is the one addition to the shared state.

Tests: a streamed thinking turn produces reasoning deltas, the reasoning item claims the first output slot, and the terminal output carries the full `thinking_blocks` payload; plus a guard that a stream without thinking gains no reasoning item. Verified live — a streamed `/v1/responses` turn against Claude now replays successfully. The `<Note>` documenting this as a limitation is removed.

Greptile's first finding, "Master key identity is misreported", is on `internal/server/auth_verify_handler.go:95`, which is not in this diff — it came in with #924. Leaving it out of scope here; worth a separate look.

All four surfaces (Messages, Messages streaming, Chat Completions, Chat Completions streaming, Responses, Responses streaming) now round-trip signed thinking blocks.

SantiagoDePolonia:
Added in 54924fe. The branch behaves correctly — the reasoning item opens, closes, and lands in the terminal output carrying `{"type":"redacted_thinking","data":"opaque"}`, with the assistant message after it — but nothing pinned that on this dialect, so the test now does.

PR #946, opened 2026-09-11
fix: map reasoning effort per provider, keep presidio placeholders unique, name provider instances in errors

SantiagoDePolonia:
Three fixes found in pre-release testing.

**Reasoning effort on OpenAI-compatible chat.** `thinking` on `/v1/messages` (and `reasoning: {effort}` on chat completions) was forwarded as `reasoning`, which OpenAI, Groq, and Fireworks reject with a 400, so Claude Code routed to those models failed as soon as thinking was on. It is now sent as `reasoning_effort`:
- OpenAI: flattened; dropped for `gpt-3.5`/`gpt-4*`/`chatgpt-*`, which reject the field.
- Groq, per model family: gpt-oss takes low/medium/high (none/minimal → low, max/xhigh → high); qwen3.8+ also takes none/default; earlier qwen3 takes none or default; other models drop it.
- Fireworks: flattened.

**Presidio placeholder collisions.** Numbering ignored placeholders already in the conversation. A `<PERSON_2>` an earlier reply carried back (a model-invented name) was reused for a new user value, merging two people and restoring the wrong value. Placeholder-shaped text in any message, response, or stream window is now reserved and never reused.

**Provider instance names in errors (#942 follow-up).** Errors and circuit-breaker messages named the provider type (`openai`) rather than the configured instance (`openai-eu`). The factory now passes the instance name to the provider HTTP clients. Env-configured providers are unchanged (name equals type); file and batch routing still use the type.

Verified live against OpenAI, Groq, Fireworks, and a Presidio analyzer.

## Summary by CodeRabbit

- **New Features**
  - Reasoning requests are translated appropriately for OpenAI, Groq, and Fireworks models.
  - Provider instance names can be customized and appear in error reporting.
  - Unsupported models receive requests without incompatible reasoning parameters.

- **Bug Fixes**
  - Prevented placeholder numbers from being reused across Presidio prompts, tool arguments, streams, and restored responses.
  - Improved handling of literal placeholders and cached responses.

- **Documentation**
  - Clarified reasoning behavior, placeholder restoration and caching, and provider names shown in circuit-breaker errors.


SantiagoDePolonia:
@coderabbitai rereview

PR #966, opened 2026-09-11
fix(passthrough): proxy provider-native error bodies, clean decode errors, cap stop sequences

SantiagoDePolonia:
Fixes three findings from the dialect e2e run (4, 7, 10), each verified against `main` first.

**Passthrough errors are provider-native again (finding 4).** `docs/features/passthrough-api.mdx` promises that provider error bodies and status codes are proxied, but every non-2xx was re-rendered in the gateway's OpenAI envelope, so the Anthropic SDK lost its typed errors and the upstream `request_id`. `/p/{provider}` (and the `/v1/messages` native-forwarding path) now relay the upstream JSON error body, status and content type untouched. Gateway-generated failures — authentication, rate limits, budgets, a provider that is not enabled, an unreachable upstream — keep the gateway's shape, as does an upstream body that is not JSON, is empty, or exceeds 1 MiB. Only the body, its media type and the headers the gateway already relayed on errors cross back, so upstream credentials, cookies and base URLs still cannot leak. Logs and the audit row keep the classified gateway error.

**Clean JSON decode errors (finding 7).** `goccy/go-json` wording reached clients verbatim: `json: slice unexpected end of JSON input` for a complete document, `cannot unmarshal number \"` for a string value, plus Go struct field names. Decode failures now name the offending member and the type it should have — `messages: must be an array`, `max_tokens: must be an integer`, `messages[1].role: must be a string` — with the member set as `param` on the OpenAI envelope. Truncated and non-JSON bodies get an honest message instead of a byte-level decoder complaint. The canonical chat/responses types wrap the nested decode error so the position that identifies the member survives their `UnmarshalJSON`.

**`stop_sequences` beyond four (finding 10).** OpenAI, Azure OpenAI and Groq cap `stop` at four items; Anthropic accepts more, and the Messages API docs say stop sequences are honored on every provider. Following Postel's law the gateway now truncates the list to each provider's limit (logging a warning) instead of letting the request fail upstream over a limit the client cannot know; Anthropic still receives the full list. The Messages API doc records the limit.

**Tested**

- `go build ./...`, `go test -race ./internal/...`, `make lint`, `make test-race` (only the pre-existing `TestVersionEndpoint*` local/UTC date flake fails).
- Live gateway against real providers: Anthropic passthrough 400/404 now return `{"type":"error",...}` with `request_id` and `anthropic-python` raises its typed `NotFoundError` with the native body; gateway auth / disabled-provider errors stay OpenAI-shaped; success bodies still byte-verbatim; malformed bodies on `/v1/chat/completions`, `/v1/messages` and `/v1/responses` return the new messages; six stop sequences now succeed on `openai/gpt-4.1-mini` and `groq/openai/gpt-oss-20b` and are still applied in full on Anthropic.

### DIFF
--- a/docs/advanced/anthropic-messages-api.mdx
+++ b/docs/advanced/anthropic-messages-api.mdx
@@ -228,7 +228,9 @@ features that have no canonical equivalent are not preserved end to end:
   `stop_reason: "stop_sequence"` plus the `stop_sequence` value. OpenAI-family
   providers conflate stop-parameter hits with natural stops in `finish_reason`,
   so completions there report `stop_reason: "end_turn"` (output is still truncated
-  correctly).
+  correctly). OpenAI, Azure OpenAI, and Groq accept at most four stop sequences;
+  a longer list is truncated to the first four (with a warning in the gateway
+  log) instead of failing the request. Anthropic accepts the full list.
 - **`thinking`** on a non-Anthropic model becomes a reasoning effort (budget
   under 10000 tokens: `low`, under 20000: `medium`, otherwise `high`; adaptive:
   `medium`), sent as each provider's own field: `reasoning_effort` on OpenAI,
--- a/docs/features/passthrough-api.mdx
+++ b/docs/features/passthrough-api.mdx
@@ -144,7 +144,10 @@ Passthrough is intentionally narrow while the API is in beta.
   that surface.
 - GoModel does not translate passthrough request bodies or response bodies.
 - Provider-native error bodies and status codes are proxied instead of converted
-  into OpenAI-compatible responses.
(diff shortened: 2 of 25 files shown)


SantiagoDePolonia:
Greptile's finding is addressed in a3d4dcb9: the Responses `input` union is decoded from its own fragment, so its decoder positions are meaningless in the request body. That failure is now tagged with the member it came from and reported as `input: must be an object` (with `param: "input"`), and the generic locator additionally discards any position whose member already holds the expected type, so no innocent member can be blamed.

SantiagoDePolonia:
@coderabbitai review

SantiagoDePolonia:
@greptileai review

SantiagoDePolonia:
@coderabbitai review

PR #976, opened 2026-09-12
test: add echotest and providertest helper packages

SantiagoDePolonia:
Adds two small test helper packages so handler and provider tests can stop repeating their setup:

- `internal/echotest`: builds an echo context and recorder from a method, target, and body (string, bytes, reader, or JSON-encoded value), with options for headers, path values, and context values, plus a generic JSON response decoder.
- `internal/providers/providertest`: recording upstream test servers (`Server`, `JSONServer`, `SSEServer`, `RouteServer`) that capture method, path, query, headers, and body, and `AssertChatCompatible`, a shared contract check for providers built on the OpenAI-compatible adapter (registration, constructors, chat, streaming, model listing, Responses translation, embeddings).

No production code changes. Follow-up PRs adopt these helpers per package and convert hand-rolled assertions to testify.

## Summary by CodeRabbit

* **Tests**
  * Added reusable helpers for constructing requests, configuring contexts, recording responses, and decoding test results.
  * Added provider test servers that capture requests, support JSON and streaming responses, and dispatch routes.
  * Added shared compatibility checks for provider registration, chat, streaming, model listing, response translation, embeddings, and unsupported capabilities.
  * Added coverage for request bodies, headers, paths, query parameters, JSON payloads, response status, and route handling.

### DIFF
--- a/internal/echotest/echotest.go
+++ b/internal/echotest/echotest.go
@@ -1,0 +1,131 @@
+// Package echotest builds echo contexts for handler unit tests so each test
+// states only what differs: method, target, body, and the occasional header
+// or path value.
+package echotest
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"maps"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/labstack/echo/v5"
+)
+
+// Option adjusts the request or context built by Request.
+type Option func(*settings)
+
+type settings struct {
+	headers     http.Header
+	pathValues  echo.PathValues
+	path        string
+	values      map[string]any
+	contentType string
+}
+
+// WithHeader sets a request header.
+func WithHeader(key, value string) Option {
+	return func(s *settings) { s.headers.Set(key, value) }
+}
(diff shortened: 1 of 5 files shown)


SantiagoDePolonia:
Addressed the review findings in the follow-up commit:

- echotest tests now cover the string, byte slice, and reader body forms verbatim.
- `AssertChatCompatible` builds the chat completion provider through `Registration.New` with the test server's base URL, so a miswired factory fails the suite.
- Every subtest now asserts the translated content: the user prompt in each chat body, the reply text in the stream and in the Responses output, the embeddings input, and the returned vector values.
- Added a `StreamResponses` subtest that checks the translated request and the `response.output_text.delta` event.
- `Capture` records body read errors and `Last` fails the test when the last request's body was not read in full.

Not applied: calling `AssertNoNativeSurfaces` from inside `AssertChatCompatible`. Some adapter-based providers legitimately expose audio or batch surfaces, so that check stays a separate opt-in call.

SantiagoDePolonia:
@coderabbitai rereview

PR #977, opened 2026-09-12
test(core): convert config, llmclient, and core tests to testify

SantiagoDePolonia:
Converts the hand-rolled `if ... { t.Fatalf }` assertions in `config/`, `internal/llmclient/`, and `internal/core/` tests to testify, and removes repeated setup.

- Adds `loadConfigYAML` in the config tests so writing a config file and loading it is one call; 19 tests use it.
- Replaces per-file JSON and pointer comparison helpers with `require.Equal`, `JSONEq`, `Same`, and `Len`.
- Removes `TestGatewayError_AsError`, which only exercised `errors.As` on a concrete type.

Test-only change, about 2,800 lines removed. Stacked on #976 (helper packages).

## Summary by CodeRabbit

- **Tests**
  - Standardized assertions across configuration, core, and client tests using clearer test helpers.
  - Preserved existing test coverage, scenarios, expected values, and application behavior.
  - Simplified repeated test setup and JSON/configuration comparisons for more consistent failure reporting.


SantiagoDePolonia:
Addressed the review findings in b0414dc0.

Fixed:
- Pointer identity checks that the testify conversion had turned into deep equality are back to `require.Same`/`assert.Same`: `pricing_timewindow_test.go` (AtTime receiver), `semantic_canonical_test.go` (cached chat/batch/file requests), `model_breaker_test.go` (retained, provider-fallback, repeat-lookup, and provider-scope breakers), `resilience_policy_test.go` (concurrent lookup keeps `assert.Same` inside the goroutine; the provider-breaker check after `wg.Wait()` uses `require.Same`). Also caught `errors_test.go` Unwrap, which was a pointer comparison on main but not flagged.
- `assert.NotNil` before a dereference is now `require.NotNil` in `config/provider_models_test.go` (context window, max output tokens, pricing, metadata override) and the same pattern in `config/config_test.go` (resilience retry, local cache dir, max price filter).

Skipped: `assert.NotNil` calls not followed by a dereference (e.g. `provider_models_test.go` mixed-list metadata, `client_test.go` response headers) were left as-is since a nil there cannot panic.

`gofmt`, `go vet`, and `go test` for `config`, `internal/llmclient`, and `internal/core` are clean; the three vet duplicate-tag warnings in `internal/core` predate this PR.

SantiagoDePolonia:
The three identity findings (pricing_timewindow_test.go, semantic_canonical_test.go, model_breaker_test.go) are already `require.Same` at the current head (ddc69745): the pointer comparisons were converted to identity assertions in the initial commit. The remaining `Equal` calls on pointer arguments in this PR compare `*string`/`*float64` values from table cases, where value equality is the intent.

PR #978, opened 2026-09-12
test(server): adopt echotest and testify in handler tests

SantiagoDePolonia:
Rewrites the `internal/server` handler tests on top of the shared `echotest` helper and testify.

- About 290 hand-built echo request/recorder/context blocks become one `echotest.Post/Get/Request` call; ~50 JSON response decodes use `echotest.Decode`.
- Per-file request builders (`setPathParam`, `newImageRequest`, the audio multipart helpers, MCP/rate-limit/session contexts) are folded into echotest or rebuilt on it.
- Hand-rolled `if ... { t.Fatalf }` assertions converted to `require`/`assert`.
- Removed four redundant tests: `TestAuthMiddleware_Integration` and `TestAuthMiddleware_ConstantTimeComparison` (branches already covered by the `TestAuthMiddleware` table, and the latter did not test timing), one duplicate table case in `TestAuthMiddleware`, and `TestConfigureGatewayHTTPServer_PreservesServerWriteTimeoutDefault` (same branch as `TestNewGatewayStartConfig_AppliesTimeoutOverrides`).

Test-only change, about 4,700 net lines removed. Stacked on #976 (helper packages).

## Summary by CodeRabbit

- **Tests**
  - Standardized server test setup with shared HTTP request and response helpers.
  - Updated assertions across authentication, audio, conversations, messaging, realtime, workflows, usage, and other server areas for clearer failure reporting.
  - Preserved existing test scenarios and expected behavior while improving consistency and maintainability.
  - Expanded structured checks for selected response, audit, streaming, header, session, and workflow outcomes.


SantiagoDePolonia:
Addressed the review findings in af6274bf:

- `request_snapshot_test.go` / `session_test.go`: the three hand-rolled `&a[0] != &b[0]` backing-array checks that the migration turned into `require.Equal` are now `require.Same`.
- `messages_handler_test.go`, `realtime_audio_alias_resolution_test.go`, `realtime_translation_test.go`, `realtime_webrtc_service_test.go` (L158), `image_edit_service_test.go`: `assert.NotNil` / `assert.Len` guards that are immediately followed by a dereference or index are now `require.*`, so a regression reports the failed assertion instead of panicking.
- `http_start_test.go`: `TestNewGatewayStartConfig_AppliesTimeoutOverrides` now seeds non-zero timeouts, restoring the override coverage lost when the old `configureGatewayHTTPServer` test was folded in.

Skipped:
- `realtime_webrtc_service_test.go` L391 (`session`) and `image_edit_service_test.go` L361 (`usage`): both are `map[string]any`; reading from a nil map does not panic, so the non-fatal assert already reports cleanly.

PR #981, opened 2026-09-12
test(providers): adopt providertest and testify in core provider tests

SantiagoDePolonia:
Rewrites the tests of the `providers` root package and the `anthropic`, `gemini`, and `openai` adapters on top of the shared `providertest` servers and testify.

- All 118 hand-written `httptest.NewServer` request-capture setups in the three adapter packages become `providertest.JSONServer`/`SSEServer`/`RouteServer` plus `capture.Last(t)`; hand-written handlers stay only where they implement multi-step behaviour (cached content reuse, key rotation, batch upload, breaker isolation).
- Hand-rolled assertions converted to `require`/`assert`; pointer identity checks use `Same`/`NotSame`, JSON comparisons use `JSONEq`.
- Redundant tests merged or removed: constructor non-nil checks in all three adapters, `TestProviderFactory_HooksPassedToBuilder` (subset of `TestProviderFactory_SetHooks`), the five per-type `TestApplyProviderEnvVars_Discovers*FromAPIKey` tests folded into one table, and the openai reasoning/GPT-5 token-parameter and tool-configuration tests folded into tables with identical per-case assertions.

The root package tests keep hand-rolled servers because `providertest` imports `providers`. Test-only change, about 6,600 net lines removed. Stacked on #976 (helper packages).


SantiagoDePolonia:
Both identity checks in registry_metadata_override_test.go now use `assert.Same` (b56823b7), along with every other site in this PR where the original test compared pointers with `!=`/`==` (cache_planner, registry, and router tests).

PR #982, opened 2026-09-12
test(admin): adopt echotest and testify in admin, auditlog, and cache tests

SantiagoDePolonia:
Rewrites the `internal/admin`, `internal/auditlog`, `internal/responsecache`, and `internal/telemetry` tests on top of the shared `echotest` helper and testify.

- Hand-built echo request/recorder/context blocks and the per-file `newHandlerContext` and decode helpers become `echotest.Get/Post/Request` and `echotest.Decode`; `echo.New()` remains only where the real router is exercised.
- Hand-rolled assertions converted to `require`/`assert`; pointer identity checks use `NotSame`.
- Redundant tests removed or merged: constructor non-nil checks and a subset test in `admin/dashboard`, `TestParseUsageParams_IntervalEmpty` (identical to an assertion in `TestParseUsageParams_DaysDefault`), the two recalculate-confirmation tests and the slash-in-name credential test folded into tables, `TestNoopLogger` (does-not-panic), `TestSQLStore_WriteBatch_Chunking` (covered by the boundary and parameter-limit tests), and the two identical cache-control skip tests merged into one.

Test-only change, about 5,800 net lines removed. Stacked on #976 (helper packages).

## Summary by CodeRabbit

* **Tests**
  * Standardized administrative, audit logging, response caching, and telemetry tests with shared HTTP test helpers and clearer assertions.
  * Preserved existing coverage for handlers, caching, audit records, provider behavior, workflows, and telemetry.
  * Consolidated several redundant or overlapping test cases while retaining expected behavior checks.


SantiagoDePolonia:
The three `assert.NotNil` before dereference sites (lines 490, 547, 587) are switched to `require.NotNil` in e9c353ac, together with the other assert-before-dereference sites in this PR.

SantiagoDePolonia:
Applied: the guardrail type test now tracks whether the `max_tokens` field was found and asserts it (178d41a8).

PR #983, opened 2026-09-12
test(providers): share the OpenAI-compatible contract across thin provider tests

SantiagoDePolonia:
Rewrites the tests of the thin provider adapters (every `internal/providers/<name>` package except anthropic, gemini, and openai) on top of `providertest` and testify.

- Fifteen providers built on the OpenAI-compatible adapter (hetzner, kilo, kimicode, meta, fireworks, deepseek, zai, xiaomi, minimax, opencodego, chutes, groq, bailian, and others) now run one `TestChatCompatibleContract` via `providertest.AssertChatCompatible`, which covers registration metadata, constructor safety, chat, streaming, model listing, Responses translation, and embeddings. The per-package copies of those tests are removed; provider-specific tests (headers, reasoning mapping, model ID rewriting, audio, images, realtime, error mapping) stay.
- Providers whose Responses path posts natively to `/responses` (oracle, sglang, vllm, llamacpp, llmd, openrouter, xai) and the azure and bedrockmantle adapters keep their own tests and are noted in the report.
- Constructor non-nil, does-not-panic, and compile-time interface tests removed; interface checks become package-level `var _` declarations. Duplicate cases merged into tables (ollama, googlecommon, chatgpt, vllm, deepseek).

Test-only change, about 5,600 net lines removed. Stacked on #976 (helper packages).

## Summary by CodeRabbit

* **Tests**
  * Standardized provider test coverage across chat, streaming, audio, embeddings, models, realtime, passthrough, authentication, routing, and error handling.
  * Expanded validation for request forwarding, response translation, streaming failures, context cancellation, and provider capability boundaries.
  * Consolidated repetitive checks into shared contract-based coverage, improving consistency across providers.

PR #991, opened 2026-09-15
test(internal): adopt testify across the remaining internal packages

SantiagoDePolonia:
Converts the remaining hand-rolled `if cond { t.Fatalf }` assertions under `internal/` (except `internal/storage`) to testify, finishing the migration started in #977–#983. Test-only; no behavior change.

- `require` before any dependent dereference, `assert` for independent checks.
- Pointer comparisons that the originals made with `==`/`!=` become `Same`/`NotSame`.
- Loop-index context that the messages carried is kept as assertion messages.

## Summary by CodeRabbit

* **Tests**
  * Standardized assertions across the test suite using structured assertion helpers.
  * Preserved existing test coverage, expected values, error handling, and behavior checks.
  * Improved failure reporting and consistency across unit, integration, storage, lifecycle, and service tests.
  * Maintained validation of success paths, failure conditions, edge cases, concurrency, persistence, and resource handling.


SantiagoDePolonia:
Applied: the recorder checks in `internal/app` use `require.Same` again, and the fetcher tests use `require.Error` before reading the message (4072c771). A type-aware scan of the branch for other reference-typed `Equal` calls also switched the plugin factory check to `NotSame` and the sentinel-error comparisons in batch and runtimesettings to `ErrorIs` (4333e7a8).

SantiagoDePolonia:
Applied in 96c4d67f: every optional pointer the modeldata, pricingoverrides, and guardrails tests dereference is now preceded by `require.NotNil`, the broker tests assert the nested map types before reading them, and the remaining `if x == nil { t.Fatal; return }` blocks became `require.NotNil`.

PR #992, opened 2026-09-15
test(server): give the third chained turn its own response ID

SantiagoDePolonia:
Fixes a flake in `TestResponsesWithPreviousResponseID_ChainCarriesFullHistory`: turn three reused the response ID `resp_conv_2`, so its asynchronous store write could overwrite turn two's snapshot before the test read it back. Locally the test failed 18 of 300 runs on main and 0 of 300 with the fix. It surfaced as a Unit Tests failure on #990, which does not touch this package.

## Summary by CodeRabbit

- **Tests**
  - Updated response-chain test coverage to use a unique identifier for each conversation turn.
  - Added clarification around asynchronous storage behavior to improve test reliability.


SantiagoDePolonia:
@coderabbitai review

SantiagoDePolonia:
Superseded: #993 landed the same fix (turn three now uses resp_conv_3) on main.

PR #1013, opened 2026-09-15
fix(virtualmodels): keep failover legs behind single-target redirects

weselben:
## TL;DR

A request through a virtual model redirect with one declared target never fails over, even when the target chains into a subtree with many alternative models. The first retry-eligible upstream error (429, 5xx, or anything matching `retry_on_statuses` / `retry_on_errors`) fails the request outright. This PR drops the declared-target-count guard in `ResolveFailovers`; the flattened leaves now decide the chain.

**Files to review (3, +96 / -1):**

| File | Why |
|---|---|
| `internal/virtualmodels/failover.go` *(start here)* | The guard removal and the doc update. |
| `internal/virtualmodels/failover_test.go` | Unit test: a single-target alias keeps the subtree legs. |
| `internal/server/failover_test.go` | Handler-level contract test with the real `virtualmodels.Service`. |

## Why

A common topology is a stable public alias with a single target that points at a routing subtree which carries the actual redundancy:

```
alias/prod   → [ router/main ]                            (1 declared target)
router/main  → [ vendor-a/model-x, vendor-b/model-y ]     (strategy: failover)
```

Requests to `router/main` fail over as designed. Requests to `alias/prod` never did.

## How

`ResolveFailovers` guarded on the declared target count of the requested redirect:

```go
if !ok || !entry.failover() || len(entry.targets) < 2 {
    return nil
}
```

For a single-target redirect this returned `nil` before `snapshot.leafTargets` ever descended the chain, so `tryFailoverResponse` / `tryFailoverStream` short-circuited on `len(failovers) == 0` and the retry policy never ran. The guard conflated "one declared target" with "one available leaf".

The fix removes the declared-target-count check. The flattened leaves are the real chain size, and the existing `seen` dedup already removes the primary resolved model. A redirect with no alternative leaf yields an empty chain exactly as before.

Behavior matrix:

| Topology | Before | After |
|---|---|---|
| 1 concrete target, resolved to it | no chain | no chain (only leaf is the primary) |
| 1 target → chain with 1 leaf | no chain | no chain (same) |
| 1 target → subtree with N leaves | no chain | N−1 legs |
| ≥2 targets | unchanged | unchanged |

## Reviewer notes

- **Native passthrough changes for these requests.** Anthropic-native requests through such an alias now take the translated path, because passthrough is off when failover legs exist. Set `failover: false` on the redirect to keep strict passthrough.
- **No change for redirects with two or more targets.** Legacy `failover.rules` migrations and self-shadowing redirects (`[self, fallback, …]`) always declare at least two targets.
- **`MaxChainDepth` is unaffected.** It is enforced at snapshot build time, not in this read path.

## Tests

- `TestFailover_SingleTargetRedirectKeepsChainedLegs` (unit): the alias returns the subtree alternative leaf as the chain.
- `TestChatCompletion_SingleTargetAliasKeepsSubtreeFailover` (handler level, real `virtualmodels.Service`): the request sweeps the subtree leg on a retry-eligible error and records primary + failover attempts in the audit entry. This pins the behavior end to end against future refactors.
- `go test ./internal/virtualmodels/ ./internal/gateway/ ./internal/server/` is green. `go test ./...` is green.

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved failover handling for aliases with a single configured target.
  * Chained failover routes now retain their remaining fallback options, allowing requests to retry alternate models when the primary model is unavailable.
  * Existing duplicate prevention and exclusion of the initially selected model remain unchanged.

---


weselben:
I can confirm it works in prod flawlessly - fixes my issue from the PR Body!

SantiagoDePolonia:
As far as I remember - to failover you need to set up at least 2 targets. First - the original model / alias - and the second - the model you want to failover to. That's why failover is load balancing strategy. When you have only one target there is nothing to load balance. AI is too dump to flag this I guess (or I'm wrong?). What do you think @coderabbitai @greptile-apps ?

SantiagoDePolonia:
@greptile-apps, @coderabbitai Speak concise and simple. When we have chain like alias → router/main → [vendor-a/model-x, vendor-b/model-y] and load balancing->failover strategy is set for router/main - would the current (pre-fix) codebase work correctly?

SantiagoDePolonia (approved):
Verified, LGTM!

PR #1025, opened 2026-09-16
perf(live): accumulate audit snapshots as typed previews

SantiagoDePolonia:
`mergeEventData` decoded both the accumulated snapshot and the incoming event and re-encoded the result, on every audit event, under the broker mutex. It accounted for about 95% of the allocations the broker charges a request — paid whether or not a dashboard is open. The active snapshot now keeps the typed preview its event was encoded from, so the next event merges onto a struct and is encoded once.

Merge semantics are unchanged: a member the patch omits (`omitempty`, so a zero field) leaves the earlier value in place, nested objects merge the same way, arrays and scalars replace. Values a snapshot keeps are copied — the preview is built from the live audit entry, which the request goes on mutating.

Three deliberate behaviour changes, each covered by tests:

- the retained-size cap now bounds the accumulated snapshot, not only the event that produced it, which is what it promises
- compaction only empties the preview's `data`, so an oversized top-level error message used to escape the cap regardless of this change (on main such an event retains 196,745 bytes against the 64KB cap). It is now trimmed, with a marker, until the event fits
- audit events published through the package-internal `publish` helper (only tests did) no longer update active snapshots; production audit events all arrive via `PublishAuditEvent`

`BenchmarkBrokerAuditRequestLifecycle`, the events one non-streaming chat request emits, n=6:

| | before | after |
|---|---|---|
| no subscribers | 44.5 µs, 80.6 KB, 1501 allocs | 7.2 µs, 14.9 KB, 89 allocs |
| one subscriber | 71.5 µs, 90.4 KB, 1529 allocs | 19.8 µs, 24.7 KB, 117 allocs |


SantiagoDePolonia:
Both reviews landed on the same finding, and it is real — though it predates this PR: `compactAuditPreviewForRetention` empties only the preview's `data`, so an oversized top-level `error_message` was never rechecked against the cap. On `main` that event retains 196,745 bytes against the 64KB limit; on this branch before the fix, 196,744.

Since this PR is what makes the cap hold for accumulated snapshots, it now also trims the top-level message after compaction, leaving a truncation marker, until the event fits. Covered at both sizes: a message just over the cap (Greptile's reproduction) and one three times it.

SantiagoDePolonia:
End-to-end numbers for this change, from the production-shape benchmark with live logs on and no dashboard connected — the default configuration of a running gateway. Same machine, n=6:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `main` | 105,400 | 160.5 KB | 2,442 |
| this branch | 29,100 | 42.5 KB | 275 |

That is roughly 76 µs, 118 KB and 2,167 allocations off every request — considerably more than the broker micro-benchmark in the description shows, because the broker runs on every lifecycle event of every request. For scale, the same request with audit logging but no broker costs 15.5 µs.

The benchmark and a guard for it are in a small PR stacked on this branch; the perf suite wires a stub audit logger today, which is why none of this was visible to a guard before.

PR #1026, opened 2026-09-16
ci(integration): take PostgreSQL from a service container

SantiagoDePolonia:
The integration job started its own PostgreSQL container from `TestMain`, after the suite had finished compiling. As a service container the runner pulls and starts it during job setup instead, so the database is ready by the time the tests are.

MongoDB still comes from `TestMain`: the suite needs a replica set, which a service container cannot be.

`GOMODEL_INTEGRATION_POSTGRES_URL` is the existing hook for an already-running database, so no test changes. It names a `_test` database because the suite drops and recreates its tables and `TestMain` refuses any other name.

## Summary by CodeRabbit

* **Tests**
  * Updated integration testing to use a consistent PostgreSQL 16 test environment.
  * Improved database connection configuration for integration tests.
  * Existing MongoDB test provisioning remains unchanged.


SantiagoDePolonia:
On pinning the PostgreSQL service image to a digest: the workflow pins every GitHub Action by commit SHA, but no container image by digest — the unit-tests job's `postgres:18-alpine`, `tools/ci/mongo-replset.sh`'s `mongo:7`, and the `postgres:16-alpine` this suite already pulls in `TestMain` are all mutable tags. Pinning this one alone would be inconsistent and would need a manual digest bump whenever the base image moves, so I've left it as a tag. Worth doing repo-wide if service images should be pinned, but that's its own change.

PR #1028, opened 2026-09-16
build: drop the brotli dependency

SantiagoDePolonia:
**`andybalholm/brotli` — removed.** The audit middleware decoded `Content-Encoding: br` on the responses the gateway itself sends. A brotli-compressed body cannot arrive there: passthrough drops a forwarded `Accept-Encoding` so Go's transport negotiates gzip itself and strips `Content-Encoding` on the way back, and the only response compression of our own is echo's Gzip on the admin routes. An encoding the decoder does not know was always passed through unchanged — exactly what happens now — so behaviour is unchanged. The `br` test case is kept, inverted, to pin the decision.

**`gopkg.in/yaml.v3` → `go.yaml.in/yaml/v3`.** The maintained fork, same API, already in the graph via testify. This drops a *direct* dependency but not the module: `swaggo/swag/v2` still requires `gopkg.in/yaml.v3`, so it stays as an indirect until D3 removes the swag runtime. Worth being precise about, since the plan claimed a removal.

**Withdrawn — replacing `golang.org/x/term`.** `os.ModeCharDevice` counts every character device as a terminal, so logs redirected to `/dev/null` would have switched from JSON to the coloured handler. That semantic change is not worth one small dependency, so `term.IsTerminal` stays. The focused test written for it stays too, now pinning the behaviour that justifies keeping it: `/dev/null` is a character device and must still log JSON.

**Withdrawn — demoting `proto/otlp` and `protobuf` to indirect.** Both are imported by the e2e-tagged OTLP collector tests in this module, so `go mod tidy` keeps them as direct requirements. It cannot be done without dropping that test coverage.

`go.mod` loses brotli outright; `x/term` stays pinned at v0.46.0. Verified with the default and `-tags=swagger` builds, the full unit suite, lint, `fix-check` and the perf guards.

## Summary by CodeRabbit

- **Bug Fixes**
  - Audit log responses encoded with Brotli are now passed through unchanged.
  - Gzip and deflate response decompression continues to work as expected.
  - Improved detection of terminal and non-terminal logging destinations for more consistent log formatting.

- **Maintenance**
  - Updated YAML configuration handling dependencies without changing configuration behavior.


SantiagoDePolonia:
Both reviews landed on the same point, and it is the one real behaviour change here: `os.ModeCharDevice` treats every character device as a terminal, so logs redirected to `/dev/null` now get the coloured handler instead of JSON.

That is deliberate — the check decides nothing but log formatting, and `LOG_FORMAT=json` overrides it either way — but it was untested, which is what Greptile asked for. `run/logging_tty_test.go` now pins all four branches: a writer that is not a file, a regular file (JSON), a character device (coloured), and a `Stat` error falling back to JSON.

If keeping `term.IsTerminal`'s exact semantics is worth the dependency, say so and I will drop the `x/term` part and keep the brotli and yaml changes — the two are independent.

SantiagoDePolonia:
Reverted the TTY change as you asked — `term.IsTerminal` stays, and so does the `golang.org/x/term` dependency. The docs wording is back to the original too.

I kept the focused test, flipped to pin the behaviour that justifies keeping the dependency: `/dev/null` is a character device but not a terminal, so redirected logs must stay JSON. That is exactly the case `os.ModeCharDevice` would have got wrong.

One thing worth noting: removing and re-adding `x/term` made `go mod tidy` resolve it to v0.45.0, a downgrade from the v0.46.0 pin. I restored `go.mod`/`go.sum` from `main` and re-tidied, so the pin is unchanged at v0.46.0.

What is left in this PR is the brotli removal and the yaml fork swap.

PR #1029, opened 2026-09-16
test(perf): guard the live broker's cost in the production shape

SantiagoDePolonia:
Stacked on #1025 — the base is that branch, so the diff here is only the benchmark and its guard.

The perf suite wires a stub audit logger, so nothing in it ever exercised the live broker. That is why neither half of P1 was visible to a guard. The audit middleware publishes lifecycle events through whichever logger implements `auditlog.LiveEventEmitter`, so this adds a bench logger forwarding to a real `live.Broker` with no dashboard connected — the default configuration of a running gateway.

Same machine, same benchmark, n=6:

| production shape, live logs on, nobody watching | ns/op | B/op | allocs/op |
|---|---|---|---|
| `main` | 105,400 | 160.5 KB | 2,442 |
| with #1025 | 29,100 | 42.5 KB | 275 |

So #1025 takes roughly 76 µs, 118 KB and 2,167 allocations off every request in this shape — considerably more than its own micro-benchmark suggested, because the broker runs on every lifecycle event of every request.

What is left is still not free. Against the stub-logger case in the same suite (15.5 µs / 20.4 KB / 160 allocs), live logs cost about 14 µs, 22 KB and 115 allocations per request with nobody watching. The guard holds that at 280 allocs / 44032 bytes so it cannot grow back unnoticed.

## Summary by CodeRabbit

- **Tests**
  - Added a production-shaped performance benchmark covering live-broker request handling.
  - Added performance guardrails for memory allocations and request size.
  - Verified successful chat responses with authentication, auditing, usage logging, sessions, rate limiting, and routing enabled.
  - Expanded performance coverage to better reflect real-world server configurations.


SantiagoDePolonia:
Rebased onto main now that #1025 has merged. The branch was stacked on #1025's branch, so after the squash merge it still carried those three commits and the diff here duplicated ~470 lines of `internal/live`. It is now just the benchmark and its guard, and it picks up the full CI suite that a non-main base was skipping.

PR #1049, opened 2026-09-18
feat: add setting to show real client IP instead of proxy IP in audit logs

iggy:
## Description

Add support for allowing GoModel to show the real client IP in audit logs. It works by extracting the first IP from x-forwarded-for headers that isn't in a new trusted cidrs setting

## Summary by CodeRabbit

- **New Features**
  - Added trusted proxy CIDR configuration for audit logging.
  - Audit entries can identify the originating client IP from `X-Forwarded-For` when requests come through configured trusted proxies.
  - Untrusted requests continue using the direct connection address; forwarding headers are ignored by default.
  - Proxy addresses are validated, normalized, and deduplicated during startup.

- **Documentation**
  - Added configuration examples and guidance covering restart requirements and security behavior.


SantiagoDePolonia:
I spoke with AI about this PR. My conclusion is that I can merge it and slightly refactor/adjust(make a few things slightly more configurable) OR if you want, you can improve it at this PR. Let me know which option do you prefer.

Mr AI told me:

> How I'd implement it
> 
> One resolver, at the server, feeding c.RealIP(). Config under server, not logging:
> 
> SERVER_TRUSTED_PROXIES=10.0.0.0/8,127.0.0.1     # CIDRs, bare IPs, or the presets below
> 
> Parse once at config load into []netip.Prefix, put it on server.Config, and build the echo.IPExtractor from it in internal/app. Then internal/auditlog/middleware.go stays exactly as it is today (ClientIP: c.RealIP()), internal/auditlog/clientip.go disappears, and every future consumer gets the same answer for free.
> 
> Use net/netip, not net.IP/net.IPNet. You're on Go 1.27. netip.Prefix/netip.Addr make roughly 60 lines of this PR vanish: canonicalProxyCIDR, singleHostCIDR, the ::ffff: mess CodeRabbit flagged, and the manual %zone stripping (Addr.Unmap(), Addr.WithZone(""), Prefix.Contains). It also stops allocating per request. The IPv4-mapped bug class CodeRabbit found only exists because of net.IPNet.
> 
> Parse once. Right now config.NormalizeTrustedProxyCIDRs validates and canonicalizes strings, then auditlog.ParseTrustedProxies re-parses the same strings with different semantics (it silently skips what the loader would have rejected). Two parsers with two behaviours for one setting is a bug waiting to happen — the comment in ParseTrustedProxies even admits it's relying on the loader having run. Have Load() produce the typed value and pass that.
> 
> Making it more configurable
> 
> The setting as shipped only handles "reverse-scan X-Forwarded-For". That misses most managed-edge deployments. I'd add, in rough priority order:
> 
> 1. Presets for the CIDR list. private (RFC1918 + ULA), loopback, cloudflare, */all. The docker/k8s case — "my ingress is somewhere in 10.0.0.0/8 and I don't know where" — is the common one, and making people hand-enumerate it is how you get 0.0.0.0/0 pasted into production. Default still empty; presets are opt-in shorthand, not implicit trust.
> 2. A configurable header. SERVER_CLIENT_IP_HEADER=X-Forwarded-For|X-Real-IP|CF-Connecting-IP|True-Client-IP|Forwarded. Behind Cloudflare or a GCP LB, the correct answer is "take CF-Connecting-IP verbatim, ignore XFF" — the chain-scan is the wrong algorithm there, not just a different header. RFC 7239 Forwarded support is a nice-to-have; the single-header mode is not.
> 3. Hop-count mode. SERVER_TRUSTED_HOPS=1 — "strip exactly N rightmost entries, take the next." This is what you want when your edge is at a fixed depth and you don't want to enumerate its (rotating) IPs. It's the mode most other gateways expose, and it composes with the CIDR mode rather than replacing it.
> 
> Point 2 is the one I'd consider a gap rather than a nice-to-have: today a Cloudflare user cannot configure this correctly at all.
> 
> 
> Smaller things
> 
> - TestMiddlewareClientIPTagging builds its context with httptest.NewRequest + echo.New().NewContext, which AGENTS.md says to do via internal/echotest. echotest has no WithRemoteAddr option yet — adding one is the right fix, and it'll be reused.
> - The docs <Note> is long and reads like a design rationale. The table row plus three sentences ("off by default; list your proxy networks; restart required") would serve users better; the reasoning belongs in the code comments, where it already is.
> - The markdown table row in configuration.mdx has misaligned pipes.

SantiagoDePolonia (commented):
Just want to make sure, that my comment 👆 was seen.

iggy:
My feelings will not be hurt if you refactor or even completely rewrite this from scratch. I'm just focused on the outcome 😆

I should say after looking through that blurb, my immediate use case is in my home lab where the clients have private IPs, but in a different range than my k8s cluster (where gomodel and my reverse proxy is running). So as long as I can only use _some_ of the private IP space CIDRs instead of all of it, it will suit my needs. It almost seemed like it was suggesting to include all private IP space by default

SantiagoDePolonia:
Thanks for expressing your use case :) It helps!

Merging this one and probably I'll modify this part a little bit soon!

SantiagoDePolonia:
@iggy FYI, the configuration shape was changed slightly here. It will be released today OR tomorrow.

That's the way it could be configured: https://github.com/ENTERPILOT/GoModel/pull/1061/changes#diff-749e06f64632f62a0c0dfbf4c4f3850e27e94ac109aa121fabd5c29469ae88de

(or just ask your AI agent about that how to set up it properly :)

PR #1053, opened 2026-09-19
feat(resilience): per-provider circuit breaker trip on quota errors with reset

weselben:
## TL;DR

Providers with usage quotas reject requests once the quota is gone, but the gateway kept sending traffic there: the observed kimicode case burned three upstream 403s per request before failover. Providers can now declare circuit-breaker trip rules: when an upstream error message matches a configured regex, the breaker opens instantly (no failure counting), requests fail fast locally, and failover serves them. After the trip TTL the existing half-open probe checks recovery, so the breaker closes itself when the real quota window resets. Operators can force-close it from the dashboard.

Behavior is unchanged without rules: zero default regexes, no env pair, no global default.

**Files to review (start here): `internal/llmclient/client_scope.go`** — single trip choke point; everything else is config, API, and UI around it.

| Area | Why |
|---|---|
| `internal/llmclient/circuit_breaker.go` | `TripRule` type, `quotaUntil` field, `RecordQuotaTrip`/`Reset`, open-with-TTL via existing open -> half-open -> closed flow |
| `internal/llmclient/client_scope.go` | Trip check in `recordCircuitBreakerCompletion` - non-streaming, streaming, embedded-200 paths all trip identically |
| `config/resilience.go` + `resilience_policy.go` | `trip_on` rule list per provider (regex + optional ttl; default = `circuit_breaker.timeout` = 30s); validation + per-provider override (list replaces global) |
| `internal/providers/credentials*.go`, `breaker_reset.go` | `ManagedProviderCredential.TripOn` (SQL JSON column + migration, Mongo BSON), `toRawProviderConfig` override pipeline |
| `internal/admin/` | `POST /admin/providers/{name}/circuit-breaker/reset` (204 / 404 / 400 / 503), status `circuit_state` + `circuit_breaker.trip_on` in `/admin/providers/status` |
| `internal/providers/openai/compatible_provider.go` + `internal/app/init_admin.go` | Chat-compatible providers expose `ResetBreaker()` -> `llmclient.Client.ResetBreaker()`; wired via `WithBreakerResetter` mirroring `WithRequestHealth` |
| `web/dashboard/src/pages/overview/` | Reset button on every provider card - clickable when breaker `open`/`half-open`, greyed when `closed` (or unconfigured). Click -> POST -> flash + status refresh |
| `web/dashboard/src/pages/providers-config/` | Rule editor for managed credentials (match + optional ttl rows, managed-only editing); read-only display for declarative providers |
| `internal/gateway/resilience_failover_test.go` | New e2e proving tripped provider A is skipped on follow-ups and re-enabled after reset (failover to B serves) |
| `internal/testconventions/assertions_test.go` | Excludes `.worktrees/` from the hand-rolled-assertion scan (pre-existing test pollution from other worktrees) |
| `docs/advanced/resilience.mdx` + `config/config.example.yaml` | Trip rule docs + commented global and per-provider examples |

## Reviewer notes

- **No behavior change without config.** No rules -> breaker behaves exactly as before. No global defaults shipped; no env pair.
- **Trip check sits only in `recordCircuitBreakerCompletion`**, so non-streaming, streaming, and embedded-200 error paths trip identically. Raw passthrough routes rely on status-based breaker behavior (documented).
- **Failover needs no new plumbing**: an open breaker already fails with 503, and 503 is in the default failover retry set.
- **Dashboard providers** edit rules in the credential editor; yaml/env providers are read-only (existing precedence rule). Managed credentials inherit global rules unless they define their own list (documented); yaml's explicit `trip_on: []` disable has no managed-UI equivalent.
- **TTL** is integer nanoseconds on the wire (Go `time.Duration` JSON); the editor stores it as a Go duration string and converts at the payload boundary. Empty TTL = use breaker timeout default (matches yaml `ttl: 0`).
- **Focus area:** TTL interplay in `acquire()` - `quotaUntil` extends the open phase past the normal breaker timeout; the half-open probe restarts the window on re-trip.

## Tests

- Unit (`internal/llmclient/circuit_breaker_trip_test.go`): trip on matching message, TTL probe recovery, zero-TTL substitution, non-matching failures keep count semantics, `ResetBreaker`, invalid regex config error, plus a DoStream matching-error trip case.
- Config (`config/resilience_policy_test.go`, `internal/providers/resilience_policy_test.go`): parsing, validation errors, per-provider override replaces global, `trip_on: []` disables, absent inherits, unset everywhere inert.
- Store round-trip (SQL + Mongo via shared `runCredentialStoreSuite`) for `ManagedProviderCredential.TripOn`.
- Admin handler tests (echotest): reset 204 / 404 / 400 / 503; credential upsert + declared-view trip_on.
- Gateway e2e (`internal/gateway/resilience_failover_test.go`): tripped breaker skips A on follow-ups; reset re-enables A.
- Dashboard (`web/dashboard/tests/`): reset button state logic (open/half-open enabled; closed/unconfigured disabled), click + refresh, editor round-trip incl. match-only row / payload ttl 0.

## Verification

`make test && make test-dashboard && make lint` green.

---

---
_This PR description was generated with AI assistance._


SantiagoDePolonia:
I need a little bit more time to think about it and review it more properly.

weselben:
@SantiagoDePolonia I also had in mind, combining it with dynamic detection for quota messages that return a refresh time (e.g. `Quota refreshes at datetime`), to have the TTL be dynamic. Maybe this whole feature could become a built-in plugin so operators can decide what to do and what is in the hot path on their own?

Regardless, reset circuit-breaker endpoint and button in UI is something I would personally like very much guess gonna make a extra PR for this later!

weselben:
**Design note: trip rules are named groups (config + env), not a list** — one more commit landed on this branch after the draft opened. Had this commit already locally but wasnt yet pushed!

@greptile @coderabbitai — semantic context, not a change request; since the maintainers want to think about this feature more deeply, here is the reasoning we ended up with.

**What changed:** `trip_on` is a docker-compose-style named map in YAML instead of a list, and env rules are declared as groups (`CIRCUIT_BREAKER_TRIP_ON_<GROUP>_MATCH` / `_TTL` globally, `<PROVIDER>_CIRCUIT_BREAKER_TRIP_ON_<GROUP>_MATCH` / `_TTL` per provider).

**Why we think it's worth it:**

1. **Env overrides become surgical.** With a list, an env rule could only replace the whole rule set — operators disabling one of the built-in rules (e.g. only the weekly quota rule) would have had to restate every other rule verbatim. With names, an env group replaces exactly the config rule of the same name (case-insensitive) and leaves the rest alone. That mirrors how this project already handles provider env overrides and rotated API keys, so the hierarchy stays: env > per-provider config > global config, applied per named group.
2. **Deterministic behavior.** A list's evaluation order is whatever the file says; a map is unordered, so the PR resolves groups to a list sorted by group name. "First matching rule wins" is now reproducible regardless of source (yaml, env, or merged), and group names double as explicit priority (e.g. kimicode ships `1_weekly_limit`, `2_five_hour_limit`, `3_usage_limit` so the specific patterns match before the catch-all).
3. **No silent failures.** A malformed group (empty pattern, invalid TTL) fails config load exactly like malformed YAML — quota protection is never dropped quietly.
4. **Operations stay stateless.** All rule shapes flow through the existing env-overlay machinery; managed providers can carry optional rule names on the wire, and the dashboard editor exposes the name as an optional field, so UI and env workflows compose instead of competing.

Happy to walk through any part of it; if a different direction (e.g. keeping the flat list) is preferred, the named-group layer is isolated enough to swap out.

> Written by AI

Issue #898, opened 2026-09-06
Issue: 'file' field required for `/v1/audio/transcriptions` - but it shouldn't be required for streaming


SantiagoDePolonia:
**Describe the bug**
`/v1/audio/transcriptions` in GoModel requires "file" - when it's not mandatory if the OpenAI allows streaming

1. Verify - is that really true for OpenAI api?
2. Implement if first is true.

ajboainain:
I found:
https://developers.openai.com/api/reference/resources/audio/subresources/transcriptions/methods/create

```
file: file
The audio file object (not file name) to transcribe, in one of these formats: flac, mp3, mp4, mpeg, mpga, m4a, ogg, wav, or webm. The request must include enough format metadata for the file to be identified. We recommend an extension-bearing filename and an appropriate content type.
```

```
stream: optional boolean or null
If set to true, the model response data will be streamed to the client as it is generated using [server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format). See the [Streaming section of the Speech-to-Text guide](https://developers.openai.com/docs/guides/speech-to-text?lang=curl#streaming-transcriptions) for more information.
```

It seems like file is required regardless of if streaming is set to true or not. I think the streaming flag is for the response- not the request being sent. Did you mean https://developers.openai.com/api/docs/guides/realtime#transcription-sessions ?

SantiagoDePolonia:
@ajboainain I think you might be right. The issue might be only related to `audio.cpp` provider (not yet in GoModel) and this anomaly should be handled in the provider specific code. 

`audio.cpp` related issue: #897

SantiagoDePolonia:
This is fixed I believe. There is a new provider `audio.cpp` - which doesn't require this field.

PR #1058, opened 2026-09-19
feat(audiocpp): add the audio.cpp provider

### LINKED ISSUE #897: Feature: New provider audio.cpp
To implement this provider: https://github.com/0xShug0/audio.cpp/

SantiagoDePolonia:
Adds `audiocpp`, a provider for [audio.cpp](https://github.com/0xShug0/audio.cpp)'s local `audiocpp_server`. Closes #897. Refs #898.

Audio-only, like ElevenLabs: `/v1/audio/speech` and `/v1/audio/transcriptions` route to it, while chat, `/responses` and embeddings return `invalid_request_error` rather than 404ing on routes the server does not have. `GET /v1/models` reports the family, mode (`tts` → `audio_speech`, `asr` → `audio_transcription`) and streaming capability audio.cpp declares per configured model; tasks with no OpenAI endpoint (alignment, diarization, …) stay listed without a mode and are reached through passthrough at `/p/audiocpp/...`, which also covers `/audio/transcriptions/details`, `/audio/transcriptions/live`, `/audio/alignments`, `/audio/voices`, `/tasks/*` and `/health`.

`AUDIOCPP_BASE_URL` is required (the server's default port collides with the gateway's own) and `AUDIOCPP_API_KEY` is optional — audio.cpp has no authentication of its own.

### `file` on `/v1/audio/transcriptions` (#898)

OpenAI does require `file` even when `stream: true`, so the general premise of the issue does not hold. audio.cpp is the case that looked like it might: it can transcribe a path on the machine it runs on, and its live route carries chunked PCM with no file at all.

Neither shape is exposed on the shared endpoint, and the gateway's own `file` requirement is unchanged. Both let the caller choose what the audio.cpp host reads or how long it holds a model, which being authorized for a model does not cover, so they stay on the native passthrough surface where the operator opts in. The same boundary applies to a `voice_ref` on `/v1/audio/speech`: inline base64 audio is forwarded, a server-side path is not. This is the conclusion for #898 rather than a code change to it — worth a maintainer decision before the issue is closed.

### Provider-specific behavior

- `voice` names an audio.cpp voice preset, cached voice id, or voice-library wav — not an OpenAI voice. Cloning fields (`reference_text`, `seed`, `options`, inline `voice_ref`, …) are forwarded verbatim.
- Synthesis returns WAV whatever `response_format` asked for, so the response is labelled from the upstream `Content-Type` rather than the requested format.
- `stream=true` transcription events are relayed with their own content type; as elsewhere in the gateway, the body is read to completion first.
- Native passthrough is **not** enabled by default: the surface is opaque, so requests there are not checked against the caller's model allowlist, and audio.cpp serves model management on it. Operators add `audiocpp` to `ENABLED_PASSTHROUGH_PROVIDERS` deliberately.

### Testing

Verified end-to-end against a real `audiocpp_server` v0.8.1 (macOS/Metal, `pocket_tts` and `moonshine_asr` tiny GGUFs): model listing, speech, transcription by upload, `response_format=text`, `stream=true`, chat refusal, the native passthrough routes including live chunked PCM transcription, and that a server-local path or a path `voice_ref` is refused on the shared routes while passthrough stays disabled by default.

## Summary by CodeRabbit

* **New Features**
  * Added support for audio.cpp as a local, audio-only provider for text-to-speech and speech-to-text.
  * Added model discovery, streaming, audio format handling, voice options, and optional authentication.
  * Added passthrough access to native audio.cpp routes, including alignment, task, and live-streaming features.
  * Added configuration examples and environment variable guidance.

* **Documentation**
  * Added setup, configuration, verification, voice, format, streaming, and endpoint documentation for audio.cpp.


SantiagoDePolonia:
Both P1 findings were valid and are fixed in 3c46c84.

**Server-local paths on the shared endpoints.** Agreed: being authorized for a model is not authorization to name a file on the audio.cpp host, and model authorization cannot cover one. `/v1/audio/transcriptions` requires an upload again, and the transport-layer change that made the `file` part optional is reverted, so the gateway behaves exactly as before for every provider. CodeRabbit's related catch on the speech route is fixed the same way: a `voice_ref` must carry inline base64 audio there, and `{"type":"path"}` or a bare path string is rejected before dispatch.

On the point that the native `/p/audiocpp/audio/transcriptions` route still forwards paths — that is deliberate and is what passthrough is: an opaque, provider-native surface that forwards the body verbatim for every provider. Parsing provider bodies to filter fields would undo that design. The boundary is the second fix instead.

**Default passthrough.** Agreed, and removed from the default list. audio.cpp's native surface includes model management (`/models/load`, `/tasks/unload_models`) and the path-based shapes, and passthrough requests carry no model selector, so it must be a deliberate opt-in. `.env.template` and the provider doc now say so and why.

**`hetzner` in the documented default list** (CodeRabbit, minor): correct, the comment did not match `config/config.go`; fixed on the line I was already touching.

**Redirect credential forwarding** (CodeRabbit, `audiocpp.go:59`): skipping. It is the shared `llmclient` redirect behavior every provider uses, not something this provider introduces, and the analysis itself concludes the exposure is bounded by the operator-configured endpoint. Worth a separate issue if the client should refuse non-loopback HTTP redirect targets generally.

Verified against a local `audiocpp_server`: a no-file request with `audio=/etc/shadow` and a path `voice_ref` are both refused with a 400 and no upstream call; normal upload transcription and speech still work; the native path shape works through passthrough once enabled; and with the default config `/p/audiocpp/models/load` is refused as a disabled passthrough provider.

PR #1067, opened 2026-09-20
feat(server): add an option to disable master key authentication

SantiagoDePolonia:
Adds `MASTER_KEY_DISABLED` (env) / `server.master_key_disabled` (YAML), default `false`, for deployments that want managed gateway API keys to be the only way in.

User-visible behavior when enabled:

- Any configured master key is forgotten at config load, so nothing accepts or derives from it — the gateway, `GET /v1/auth/verify`, and the version-check identity. A key left in the environment or `config.yaml` is inert rather than a startup error, so the switch can be flipped before the value is removed from the host.
- It fails closed, unlike an empty `GOMODEL_MASTER_KEY`. The auth middleware is always registered, every request needs a credential instead of falling back to unauthenticated access, and the anonymous `/admin/*` bootstrap bypass stays closed so the admin API requires a managed key with dashboard access.
- With no managed keys and no authentication extension, every request is rejected; startup logs a warning saying so and how to recover.

The flag is negative so the zero value of a `config.Config` literal keeps today's behavior.

Docs: new "Disabling the master key" section in `docs/advanced/configuration.mdx`, plus `.env.template` and `config/config.example.yaml`.

## Summary by CodeRabbit

* **New Features**
  * Added an option to disable master-key authentication through environment or YAML configuration.
  * When enabled, requests require valid managed gateway API keys, including access to admin and authentication verification endpoints.
  * Configured master keys are ignored and cannot authenticate requests.
  * Added startup messaging and warnings when no usable authentication credentials are configured.

* **Documentation**
  * Added guidance for enabling master-key-disabled mode and explained its behavior when no managed keys are available.


SantiagoDePolonia:
Fixed: the docs, `.env.template`, `config.example.yaml`, and the `MasterKeyDisabled` field comment now say that managed keys **and any configured authentication extension** remain valid credentials, instead of implying managed keys are the only ones. The behavior was already correct — `RequestAuthenticators` are untouched by the switch, and the startup warning about a locked-out gateway only fires when there is no managed key *and* no extension auth.

PR #1068, opened 2026-09-21
feat(minimax): request reasoning_split from the MiniMax API

### LINKED ISSUE #1032: minimax: parse inline <think> blocks into reasoning_content (Groq reasoning parity)
## Summary

MiniMax reasoning models (MiniMax-M3, MiniMax-M2.7) emit chain-of-thought as inline `<think>...</think>` text inside the message content. The gateway forwards this text as content. Clients that render reasoning from the `reasoning_content` field (for example Kimi Code CLI) show the raw XML in the chat and cannot display the reasoning.

Request: parse `<think>` blocks in the MiniMax provider adapter and return them as `reasoning_content`, the same way the Groq adapter already normalizes reasoning.

## Current behavior

- MiniMax M3 emits thinking inline. M3 supports adaptive thinking; M2.7 always thinks.
- `internal/providers/minimax/` has no reasoning handling. The `<think>` text reaches the client inside `choices[].message.content` and inside streamed `content` deltas.
- A client on the OpenAI protocol reads `reasoning_content`. It renders the XML as normal answer text instead.

## Expected behavior

- Non-stream: the adapter strips the `<think>` tags and moves the inner text to `choices[].message.reasoning_content`. Content keeps only the answer.
- Stream: while a `<think>` block is open, deltas go out as `reasoning_content` deltas; after the closing tag, deltas go out as `content`. The tags never reach the client.
- No behavior change for responses without `<think>` blocks.

## Precedent in this repo

The Groq adapter already does this class of normalization:

- `internal/providers/groq/reasoning.go`
- `internal/providers/groq/reasoning_response.go`

Per the provider overview docs: on the gpt-oss and qwen3 families GoModel defaults `reasoning_format` to `parsed`, so chain of thought arrives in `reasoning` instead of as inline `<think>` text, and Groq's `reasoning` is renamed to `reasoning_content`. The same overview states the goal: reasoning text comes back as `reasoning_content` on every provider.

MiniMax is the gap. It ships reasoning as XML in content, and the adapter passes it through untouched.

## Why guardrails cannot close the gap

Guardrail plugins (`string_replace`, `llm_based_altering`) edit message text only. No plugin can write the `reasoning_content` field. A response-phase `string_replace` can delete the block, but then the client loses the reasoning. Provider-adapter parsing is the only place that can re-label the same tokens as reasoning.

## Suggested design

- Add a parse step to `internal/providers/minimax` for message and delta content, mirroring the Groq reasoning files.
- Guard it with a provider or model option, default on for thinking models (M3, M2.7), with an opt-out for raw passthrough, mirroring Groq's `reasoning_format`.
- Keep `/v1/chat/completions` and `/v1/responses` translations consistent.

## Edge cases to cover

1. Content before and after a think block survives verbatim.
2. Multiple think blocks in one completion.
3. Unterminated block (`finish_reason: length`): treat the remainder as reasoning or content, pick one, document it.
4. Tool calls interleaved with thinking deltas.
5. Empty think block.
6. Usage accounting: reasoning tokens stay completion tokens; no double count.
7. Vendor passthrough members stay untouched (ADR-0011 forwarding).

---

- Affected docs: `docs/providers/minimax.mdx` (no reasoning section today).

> Written with AI

weselben:
## TL;DR

MiniMax's OpenAI-compatible API moves the chain of thought out of `content` and into `reasoning_content` / `reasoning_details` when the request sets `reasoning_split: true`. This PR sets that flag for the known reasoning models and normalizes the response into GoModel's canonical `reasoning_content` member — no inline tag parsing at the gateway. Supersedes the parsing approach discussed in #1041; closes #1032.

## Files to review (5, +815 / -4)

| File | Why |
|---|---|
| `internal/providers/minimax/reasoning.go` *(start here)* | Model gate, request adaptation (`reasoning_split: true` via `MergeUnknownJSONFields`, mirroring the Groq `reasoning_format` pattern), buffered normalization: joins `reasoning_details[].text` into `reasoning_content`. |
| `internal/providers/minimax/reasoning_stream.go` *(new)* | SSE rewriting: `delta.reasoning_details[].text` is cumulative upstream, so a per-choice buffer emits only the suffix as `delta.reasoning_content`; `reasoning_details` is stripped from output; MiniMax's trailing full-message summary event folds into a usage-only chunk. |
| `internal/providers/minimax/minimax.go` | Wires `adaptChatRequest` into `ChatCompletion` and `StreamChatCompletion`; `Responses` inherits through `ResponsesViaChat`. |
| `internal/providers/minimax/reasoning_test.go` *(new)* | Gate table, wire-level request capture via providertest, buffered normalization cases. |
| `internal/providers/minimax/reasoning_stream_test.go` *(new)* | Cumulative diff, passthrough, summary-event folding, big-int precision, non-gated byte-for-byte passthrough. |

## Gate

The adaptation runs only for an exact, case-insensitive allowlist: `minimax-m2`, `minimax-m2.5`, `minimax-m2.5-highspeed`, `minimax-m2.7`, `minimax-m2.7-highspeed`, `minimax-m3`. Everything else passes through untouched, byte for byte. An existing `reasoning_split` value on the request is never overwritten.

## Behavior

- **Buffered.** `reasoning_details[].text` concatenates into `reasoning_content`; an upstream `reasoning_content` wins. `reasoning_details` stays on the message so clients can send it back — the upstream requires it in history to preserve the reasoning chain across tool-call rounds.
- **Streaming.** `reasoning_details` arrives cumulative; the gateway diffs it and emits incremental `reasoning_content` deltas, matching what every OpenAI-compatible client expects. MiniMax's trailing full-`chat.completion` summary event (real `usage`, duplicated message) folds into a usage-only chunk.
- **Unknown models.** Not on the allowlist, no flag set, no normalization — zero hot-path cost.

Out of scope: vLLM-hosted MiniMax M2 (no `reasoning_split` support there), `thinking.type` control.

## Reviewer notes

- **Why not parse the tags?** Superseded by the native knob, per maintainer discussion on #1032. The parsing approach lived in #1041 (closed, kept as fallback reference).
- **Focus area:** the cumulative-diff buffer in `reasoning_stream.go` — upstream's `reasoning_details[].text` is a running total, not a delta. Covered by a two-chunk test asserting suffix-only emission.
- **Envelope safety.** Raw JSON members are re-emitted untouched; big-int precision pinned by test.

## Verification

- `gofmt` clean, `golangci-lint` 0 issues, `go build ./...` clean.
- `go test ./internal/providers/minimax/... -count=1` green; 100% coverage on `reasoning.go` and `reasoning_stream.go`.

---

## Summary by CodeRabbit

- **New Features**
  - Added support for MiniMax reasoning models, including M2, M2.5, M2.7, M3, and high-speed variants.
  - Reasoning output is available through `reasoning_content` in standard and streaming responses.
  - Reasoning requests automatically enable split reasoning when appropriate while respecting caller settings.
  - Redundant reasoning metadata is removed from streamed responses.
  - Non-reasoning MiniMax models continue to pass through unchanged.

- **Documentation**
  - Added guidance for using MiniMax reasoning models and interpreting their responses.


weselben:
Round on 6794ea39 — patch coverage gap closed.

The two uncovered lines were the upstream-error returns in `ChatCompletion` and `StreamChatCompletion`. A 500-response test now asserts the error propagates from both entry points. `minimax.go` is back at 100% function coverage.

weselben:
Opening review round for this PR.

@greptile review
@coderabbitai review
> Written by AI

weselben (commented):
Self-review round (pr-review skill loop): 6 findings — 3 yellow, 2 nits, 1 question. Triage inline: 2 get fixed now (summary collapse envelope, SSE data: prefix), 1 documented (docs section), 2 skipped with reasons (details retention is the upstream multi-turn contract; marker over-match is relay-safe), 1 answered (allowlist is a deliberate exact list).

Validation already green: build, lint, tests, 100% coverage on both new files.

(docs finding: the Reasoning section for docs/providers/minimax.mdx lands with the fix commit — the file is not part of this diff, so it had no anchor for an inline comment.)

internal/providers/minimax/reasoning_stream.go
```diff
@@ -0,0 +1,191 @@
...
+		if _, isMessage := choice["message"]; isMessage {
+			// MiniMax M3 ends a reasoning stream with a full chat.completion
+			// object — the complete message plus the real usage. Clients have
+			// already received every delta, so the event collapses to an empty
+			// choices array plus the usage and the duplicate message goes.
+			usage, hasUsage := chunk["usage"]
+			if !hasUsage {
+				return chunk, false, nil
+			}
+			return map[string]json.RawMessage{
```
weselben: 🟡 Valid — fixing. The collapse will start from the decoded chunk (keeping id/object/created/model), empty the choices array, and hoist a present finish_reason onto the emptied choice so recordFinishReason still fires on translated Responses streams.

internal/providers/minimax/reasoning.go
```diff
@@ -0,0 +1,116 @@
...
+	for i := range resp.Choices {
+		normalizeReasoningMessage(&resp.Choices[i].Message)
+	}
+}
+
+// normalizeReasoningMessage concatenates a message's reasoning_details texts
+// into reasoning_content. A message that already carries reasoning_content
+// wins, and a reasoning_details value of the wrong shape leaves the message
+// untouched.
+func normalizeReasoningMessage(msg *core.ResponseMessage) {
```
weselben: 🟡 Skipping, deliberate: upstream requires reasoning_details preserved in multi-turn history (tool-call rounds), so buffered keeps both members. The stream path normalizes to reasoning_content instead — divergence is documented in the docs section added with the next commit.

internal/providers/minimax/reasoning_stream.go
```diff
@@ -0,0 +1,191 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/goccy/go-json"
+)
+
+// sseDataPrefix introduces the JSON payload of an SSE event.
+var sseDataPrefix = []byte("data: ")
```
weselben, line 12: 🔵 Valid — fixing. Optional-space trim after the colon before the marker check.

internal/providers/minimax/reasoning_stream.go
```diff
@@ -0,0 +1,191 @@
...
+
+// Markers gate the per-chunk decode: content deltas, role chunks and the
+// terminal [DONE] line never carry them and are relayed with their original
+// bytes, so only reasoning turns pay for a re-encode. The summary marker
+// catches the trailing chat.completion event, whose full message announces
+// the end of an M3 reasoning stream.
+var (
```
weselben: 🔵 Skipping: over-match only costs a wasted decode on error-heavy streams; both shapes relay unchanged. Not worth the extra guard.

internal/providers/minimax/reasoning.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/enterpilot/gomodel/internal/core"
+)
+
+// MiniMax's reasoning models keep the answer content clean only when the
+// request asks them to split the chain of thought out of it: with
+// reasoning_split set, the reasoning arrives in reasoning_details and
+// content carries just the answer.
+const (
+	reasoningSplitKey     = "reasoning_split"
+	reasoningDetailsKey   = "reasoning_details"
```
weselben: ❓ Answered: exact allowlist is the accepted cost. Unknown members make MiniMax reject the whole request, so a broad match would break non-reasoning future models; config-driven extension is out of scope for this PR.

SantiagoDePolonia:
My AI said:

> Thanks for the rewrite, the request side looks good. Before I review the response handling, could you post raw captures from the live API with `reasoning_split: true`, one non-streaming and one streaming (`curl -sN`, unedited apart from the key)?
> 
> Most of this PR (~290 of the non-test lines) handles the response, and I can't tell from the docs whether it's needed. They say in one place that thinking comes back in `reasoning_details`, and in another that it's "exposed through `reasoning_content` and `reasoning_details`". If upstream already sends `reasoning_content`, the PR's own "already canonical" checks make `normalizeChatResponse`, `rewriteChunk` and `reasoningSuffix` no-ops, and we can reduce this to `adaptChatRequest` plus the wire-capture test.
> 
> The captures should also settle:
> 
> 1. **Cumulative deltas:** is `delta.reasoning_details[].text` really a running total on every chunk?
> 2. **Trailing summary event:** I can't find the final full `chat.completion` event in the docs. If it doesn't show up in the capture, let's drop the `"message"` marker and the usage folding.
> 3. **Allowlist:** the docs list `MiniMax-M2.1` and `MiniMax-M2.1-highspeed` as supported, and both are missing here. Did you actually see a non-reasoning model reject `reasoning_split`? If they just ignore it, I'd rather send it on every request (as suggested in #1032) than maintain a list we have to touch for each new model.
> 4. **Defensive branches:** `choiceIndex` with its fallback, the "buffer replaced" case and the per-choice map. Unless the capture shows these happening, please remove them.
> 
> Please add whatever the captures show as test fixtures, so the tests pin real upstream behavior and not hand-written chunks.

internal/providers/minimax/reasoning_stream.go
weselben: Fixed in 7c544ef2 — the collapse now starts from the decoded chunk: `id`/`object`/`created`/`model` survive, and the choices array keeps one emptied choice carrying `index` plus a hoisted `finish_reason`, so `recordFinishReason` still fires downstream.

_Written by AI._

internal/providers/minimax/reasoning.go
weselben: Docs landed in 7c544ef2 — new Reasoning section in docs/providers/minimax.mdx covers the gated models, the `reasoning_split` flag, and the normalization contract, including the buffered/stream divergence noted above.

_Written by AI._

internal/providers/minimax/reasoning_stream.go
weselben: Fixed in 7c544ef2 — the marker check now accepts `data:` with or without the space; a bare-field line reaches the rewrite instead of leaking cumulative reasoning_details.

_Written by AI._

weselben:
Round on 7c544ef2 — self-review fixes landed.

- Summary collapse keeps the chunk envelope (`id`, `object`, `created`, `model`) and hoists `finish_reason` onto the emptied choice, so completion state reaches Responses translation.
- SSE `data:` fields without the space now reach the rewrite instead of leaking raw cumulative details.
- New "Reasoning" docs section: gated models, the flag, the normalization contract, and the buffered/stream divergence on `reasoning_details` (kept buffered — upstream multi-turn contract — normalized away on stream).

Verification: gofmt clean, golangci-lint 0 issues, tests green, `go build ./...` clean, 100% coverage on `reasoning.go` and `reasoning_stream.go`.

_Written by AI._

weselben:
@greptile review
@coderabbitai review

weselben:
> Most of this PR (~290 of the non-test lines) handles the response, and I can't tell from the docs whether it's needed.

You were right to push on this. Captures pulled, response machinery gone.

**Captures** (raw, `reasoning_split: true`, unedited apart from nothing sensitive):
- non-streaming: `message.reasoning_content` (plain string) AND `message.reasoning_details` arrive natively, `content` is clean
- streaming: `delta.reasoning_content` and `delta.reasoning_details[].text` are both **incremental** — chunk 1 `"The"`, chunk 2 `" user wants a proof…"` — not cumulative as the docs example suggested
- stream ends with a standard usage-only chunk (`"choices":[]` + real usage, `base_resp`), **no trailing full-message summary event**
- `finish_reason` rides the last content chunk

**What that killed (commit 8d0a8c55, −385/+167):**
- the cumulative-diff buffer — it was built for cumulative semantics that don't exist, and would have corrupted real output
- the trailing-summary fold and the `"message"` marker — no such event
- buffered `normalizeChatResponse` entirely — upstream `reasoning_content` is already canonical
- finish-reason hoisting — `finish_reason` already rides the last content chunk

**What stays:** `adaptChatRequest` (the `reasoning_split: true` gate), and a 40-line stream strip that deletes the duplicated `reasoning_details` member from deltas so clients only see canonical `reasoning_content`. Test fixtures are the verbatim capture lines.

**Your remaining points:**
- Allowlist: kept the exact list (M2/M2.5/M2.7/M3 + M2.x-highspeed) because we cannot prove non-reasoning models merely ignore the flag — a rejection would kill whole requests. Happy to flip to "send always" with a follow-up if you prefer.
- Defensive branches (`choiceIndex` fallback, buffer-replaced case, per-choice map): all removed — the capture shows single-choice incremental chunks.
- Fixtures: tests now pin the verbatim capture lines, not hand-written chunks.

_Written by AI._

weselben:
Response handling rewritten against live captures — please re-review.

@greptile review
@coderabbitai review

internal/providers/minimax/reasoning_stream.go
```diff
@@ -0,0 +1,191 @@
...
+			// already received every delta, so the event collapses to an empty
+			// choices array plus the usage and the duplicate message goes.
+			usage, hasUsage := chunk["usage"]
+			if !hasUsage {
+				return chunk, false, nil
+			}
+			return map[string]json.RawMessage{
+				"choices": json.RawMessage(`[]`),
+				"usage":   usage,
+			}, true, nil
```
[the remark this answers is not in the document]
weselben: Stale after 8d0a8c55 — the terminal chunk is no longer replaced at all. The summary-collapse branch was deleted when the live captures showed MiniMax ends streams with a standard usage-only chunk; it now relays byte for byte with its full envelope, so id/object/model reach the usage observer untouched.

_Written by AI._

docs/providers/minimax.mdx
```diff
@@ -50,6 +50,15 @@ MiniMax requires `temperature` in `(0.0, 1.0]` and rejects zero. GoModel clamps
 a zero or negative temperature to `1.0` so OpenAI-style requests that pin
 `temperature: 0` keep working.
 
+## Reasoning
+
+`MiniMax-M3`, `MiniMax-M2.7`, `MiniMax-M2.5` and `MiniMax-M2` (plus the M2.x
+`-highspeed` variants) write their chain of thought separately from the answer.
```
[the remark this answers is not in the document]
weselben: Resolved in 563ae5bd — the Reasoning section now lists `minimax-m2.5-highspeed` and `minimax-m2.7-highspeed` explicitly instead of the loose `M2.x -highspeed variants` phrase.

_Written by AI._

weselben:
Round on 563ae5bd — docs wording only.

- The Outside-Diff "summary choice" item no longer applies: the summary-collapse branch was removed in 8d0a8c55 after the live captures showed no such event; every choice now flows through one in-place strip loop that is multi-choice safe by construction.
- Greptile's P2 (terminal event identity) is likewise stale — the terminal chunk relays byte for byte with its full envelope.
- CodeRabbit docs minor fixed: the two supported highspeed model IDs are now listed explicitly.

Verification: docs-only change; package tests green, build clean.

_Written by AI._

weselben:
Docs wording round pushed.

@greptile review
@coderabbitai review

weselben:
On the Outside-Diff item from the 16:27 review (`reasoning_stream.go:118-140`, "Preserve every terminal summary choice"): no longer applicable.

That review ran against 7c544ef2. Commit 8d0a8c55 removed the summary-collapse branch entirely after live captures showed MiniMax ends reasoning streams with a standard usage-only chunk — no full-message terminal event exists. The stream path is now a single in-place strip loop that iterates every choice, so a hypothetical multi-choice summary is handled uniformly; there is no rebuild-and-return branch left to preserve. The docs wording from the same review landed separately in 563ae5bd.

_Written by AI._

docs/providers/minimax.mdx
```diff
@@ -50,6 +50,18 @@ MiniMax requires `temperature` in `(0.0, 1.0]` and rejects zero. GoModel clamps
 a zero or negative temperature to `1.0` so OpenAI-style requests that pin
 `temperature: 0` keep working.
 
+## Reasoning
+
+`MiniMax-M3`, `MiniMax-M2.7`, `MiniMax-M2.7-highspeed`, `MiniMax-M2.5`,
+`MiniMax-M2.5-highspeed` and `MiniMax-M2` write their chain of thought
```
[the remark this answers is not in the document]
weselben: Fixed in 9c0dfcb5 — the Reasoning section now states the flag is defaulted to `true` only when the request does not already carry it, so a caller-provided `reasoning_split: false` stays in charge.

_Written by AI._

weselben:
Round on 9c0dfcb5 — docs only.

Documented the caller override: `reasoning_split` is defaulted to `true` only when absent; a caller-provided `false` is never overwritten. Everything else unchanged — tests, coverage, build all green.

_Written by AI._

weselben:
Docs override note pushed.

@coderabbitai review

weselben:
@coderabbitai full review

_Final full review before mark-as-ready, on commit `9c0dfcb5`._

_Written by AI._

weselben:
@greptile review

_Final review on commit `9c0dfcb5` — since your 5/5 score on `8d0a8c55` only docs commits landed, please take another look against the current HEAD._

_Written by AI._

internal/providers/minimax/reasoning_stream.go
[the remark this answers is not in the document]
weselben: **Stale at HEAD — resolved by `8d0a8c55`.**

This P2 describes the implementation reviewed at `6794ea39`, where the trailing summary event was collapsed into a newly built chunk envelope (`usage` copied into a fresh map), dropping upstream `id`, `object`, and `model`. That response rewriting was removed entirely in `8d0a8c55` after live stream captures showed the provider already sends the canonical shape.

On the current HEAD (`9c0dfcb5`) the terminal usage-only chunk carries no `reasoning_details` marker and is relayed byte for byte, and for rewritten chunks only `choices` is replaced while every other member keeps its original raw JSON. Provider response identity is preserved end to end.

_Written by AI._

weselben:
Deleted an accidental CR Comment i triggered!

weselben:
@coderabbitai resume

weselben:
**Resolved in `7f4eb7da` — details-only reasoning deltas are now relayed unchanged.**

The full-review finding was valid: `stripDeltaReasoningDetails` removed `reasoning_details` even when the same delta carried no `reasoning_content`, and a caller-set `reasoning_split` still routes that stream through the strip pass — so a details-only delta lost its only copy of the reasoning.

Change: `reasoning_details` is now stripped only when the same delta also carries `reasoning_content`; everything else is relayed byte for byte.

- Test "a delta with reasoning_details but no reasoning_content is relayed byte for byte" asserts the passthrough.
- New entry-point test `TestStreamChatCompletion_CallerSplitFalseRelaysDetailsOnlyDeltas` sends `reasoning_split: false` upstream and asserts the details-only delta arrives untouched and the caller flag reaches the provider unchanged.
- Gates: gofmt clean, `golangci-lint` 0 issues, package tests green, `go build ./...` clean, 100% statement coverage on `reasoning.go` and `reasoning_stream.go`.

_Written by AI._

weselben:
@greptile review

_HEAD moved to `7f4eb7da` — one more fix landed after your last pass (details-only reasoning deltas are now relayed unchanged). Please review the current state._

_Written by AI._

weselben:
@coderabbitai review

_Heads-up: HEAD moved to `7f4eb7da` 44 minutes ago (details-only delta fix) and no incremental review has landed since. Please take a look at the current state._

_Written by AI._

weselben:
@greptile review

_Reminder: HEAD is now `7f4eb7da` — the details-only delta fix landed after your last pass. Please review the current state when you get a chance._

_Written by AI._

weselben:
@coderabbitai full review

_Your last completed full review ran on `9c0dfcb5`; the incremental reply above said "Already reviewed the last commit" for `7f4eb7da`, but no review of that commit exists on the PR. Please run the full review against the current HEAD `7f4eb7da`._

_Written by AI._

SantiagoDePolonia (approved):
LGTM

weselben:
I'm going to test this in Docker once I get back into my office! After manual confirmation, I'm going to merge.

SantiagoDePolonia:
> I'm going to test this in Docker once I get back into my office! After manual confirmation, I'm going to merge.

Sounds good! Fingers crossed!

weselben:
**Verified `7f4eb7da` end to end against a live MiniMax-M2.7 deployment.**

Built this PR in Docker and pointed it at a live OpenAI-compatible MiniMax endpoint. What the PR now does:

- **Request side:** Every gateway model on the reasoning allowlist gets `reasoning_split: true` merged into the upstream request, unless the caller set the flag themselves. An explicit `reasoning_split: false` reaches MiniMax unchanged, and the inline output is relayed verbatim.
- **Response side (buffered):** With the flag set, MiniMax keeps `content` clean and delivers reasoning natively in `reasoning_content`. Without the flag, inline `<think>…</think>` blocks still arrive in `content` today — the issue premise holds upstream.
- **Response side (streaming):** The gateway drops the redundant `reasoning_details` member (which duplicates `reasoning_content` per delta) from every delta and relays everything else — including the usage-only terminal chunk — byte for byte.

Verified all four: default setter fires, caller override is honored, `content` stays clean, and `reasoning_details` never reaches the client.

_Written by AI._

